PR #67, opened 2026-03-22
#66 Serve Tesla public key at .well-known endpoint

### LINKED ISSUE #66: Serve Tesla public key at .well-known endpoint and configure fleet telemetry
Serve the Tesla app public key at `/.well-known/appspecific/com.tesla.3p.public-key.pem` on the client port so Tesla can verify our app identity during fleet_telemetry_config push.

## Tasks
- [ ] Add handler for `GET /.well-known/appspecific/com.tesla.3p.public-key.pem` on client port
- [ ] Read public key from `TESLA_PUBLIC_KEY` env var
- [ ] Add to Railway env vars: `TESLA_PUBLIC_KEY`, `TESLA_PRIVATE_KEY`

## Context
The virtual key is already paired via the my-robo-taxi Next.js app. The same keypair must be used. The public key is currently served from myrobotaxi.app (Vercel) — it also needs to be served from telemetry.myrobotaxi.app for the fleet telemetry config push.

tnando:
Tesla Fleet Telemetry requires the app public key at `/.well-known/appspecific/com.tesla.3p.public-key.pem`. This endpoint reads from the `TESLA_PUBLIC_KEY` env var.

Closes #66

### DIFF
--- a/cmd/telemetry-server/main.go
+++ b/cmd/telemetry-server/main.go
@@ -163,8 +163,7 @@ func run() error { //nolint:funlen // composition root — sequential dependency
 	}
 
 	// --- HTTP servers ---
-	teslaPublicKey := os.Getenv("TESLA_PUBLIC_KEY")
-	srv := server.New(cfg.Server(), logger, db, reg, teslaPublicKey)
+	srv := server.New(cfg.Server(), logger, db, reg, cfg.TeslaPublicKey())
 	srv.SetTeslaHandler(recv.Handler())
 	originPatterns := cfg.WebSocket().AllowedOrigins
 	if len(originPatterns) == 0 {
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -9,14 +9,15 @@ import "time"
 // Config holds the fully-validated, immutable application configuration.
 // All access is through getter methods — there are no exported setters.
 type Config struct {
-	server      ServerConfig
-	tls         TLSConfig
-	database    DatabaseConfig
-	telemetry   TelemetryConfig
-	drives      DrivesConfig
-	websocket   WebSocketConfig
-	auth        AuthConfig
-	mapboxToken string
+	server         ServerConfig
+	tls            TLSConfig
+	database       DatabaseConfig
+	telemetry      TelemetryConfig
+	drives         DrivesConfig
+	websocket      WebSocketConfig
+	auth           AuthConfig
+	mapboxToken    string
+	teslaPublicKey string
 }
 
 // ServerConfig holds port bindings for the three HTTP listeners.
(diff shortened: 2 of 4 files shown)


tnando:
**Claude finished @tnando's task in 1m 4s** —— [View job](https://github.com/tnando/my-robo-taxi-telemetry/actions/runs/23414305438)

---
### PR Review: #66 Serve Tesla public key at .well-known endpoint

**Verdict: REQUEST_CHANGES**

**Domain focus:** `agent:go-engineer`, `agent:tesla-telemetry`

---

### Findings

#### Critical (must fix)

1. **🐛 `SetClientHandler` overwrites the `.well-known` route** — This is a functional bug. The new endpoint is registered on `clientMux` in `New()` (line 65-71), but `SetClientHandler()` (line 114-120) creates a **brand new mux** that only includes `/healthz` + the provided handler. In production, `main.go:172` calls `srv.SetClientHandler(...)`, which **completely replaces** the client server's handler — the `.well-known` endpoint is silently lost and will never be reachable.

   [Fix this →](https://claude.ai/code?q=In%20internal%2Fserver%2Fserver.go%2C%20the%20SetClientHandler%20method%20creates%20a%20new%20mux%20that%20drops%20the%20.well-known%20route%20registered%20in%20New().%20Fix%20SetClientHandler%20to%20also%20re-register%20the%20.well-known%20endpoint%2C%20or%20better%20yet%2C%20store%20the%20public%20key%20on%20the%20Server%20struct%20and%20register%20all%20client%20routes%20in%20one%20place.&repo=tnando/my-robo-taxi-telemetry)

#### Warnings (should fix)

2. **Config pattern violation** — CLAUDE.md says *"Use a single `Config` struct loaded from environment variables"* and *"Validate all config at startup — fail fast on invalid config."* Reading `os.Getenv("TESLA_PUBLIC_KEY")` directly inside `New()` bypasses the config system. This should be a field on `config.ServerConfig` (or a sibling config struct), loaded and validated at startup alongside everything else.

3. **Import ordering** — `"os"` is placed between `"net"` and `"net/http"`, breaking alphabetical stdlib ordering. Should be: `"net"`, `"net/http"`, `"os"`.

4. **No tests** — There are no tests for the new endpoint (happy path: key present; edge: key absent). Given that the route gets silently dropped by `SetClientHandler`, a test would have caught the critical bug above.

#### Suggestions

5. **Consider which server should host this** — The `.well-known` path is for Tesla's fleet configuration process, not for browser clients. It may belong on the **metrics/public** server or a dedicated admin mux rather than the client mux. Worth a quick decision — Tesla's docs say it needs to be publicly accessible at the domain root, so client port may actually be correct, but worth documenting the reasoning.

6. **Silent no-op when env var is missing** — If `TESLA_PUBLIC_KEY` is empty, the endpoint simply doesn't get registered with no warning. Consider logging a warning so operators know the endpoint is inactive. In production this could cause a confusing Tesla integration failure.

---

### Domain-Specific Notes

**`agent:go-engineer`:**
- The handler closure itself is clean and idiomatic ✅
- Good use of `Cache-Control` header for a static key ✅
- `fmt.Fprint` is fine here since the PEM is a simple string ✅

**`agent:tesla-telemetry`:**
- Correct `.well-known` path (`/.well-known/appspecific/com.tesla.3p.public-key.pem`) per Tesla's Fleet Telemetry docs ✅
- Correct `Content-Type: application/x-pem-file` ✅
- No private key exposure risk — this is the **public** key, which is safe to serve ✅
- The key is loaded from an env var (not a file path), which is correct per CLAUDE.md security rules ✅

PR #112, opened 2026-03-28
Fix driving/parked status flicker in real-time UI

tnando:
## Summary

Fixes the bottom sheet flickering between driving and parked states during live driving.

**Root cause:** `deriveVehicleStatus()` ran on EVERY WebSocket update, but Tesla only sends `gear`/`speed` fields when their values change. Updates without those fields defaulted to `status: "parked"`, which the frontend merged over the existing `"driving"` status. The UI then flipped back to driving on the next update that included gear.

**Fix:** Only inject `status` into the update when `gearPosition` or `speed` is present. Updates without these fields no longer override the frontend's current status.

This also fixes the app crash — the rapid driving↔parked oscillation caused the map, bottom sheet, and all child components to remount on every update cycle.

## Test plan
- [x] `go vet`, `golangci-lint`, `go test`, `go build` — all pass
- [ ] Drive test: UI should stay stable in driving mode

### DIFF
--- a/internal/ws/broadcaster.go
+++ b/internal/ws/broadcaster.go
@@ -116,9 +116,16 @@ func (b *Broadcaster) handleTelemetry(ctx context.Context, event events.Event) {
 	// ordering) — lastUpdated is what the UI displays.
 	fields["lastUpdated"] = payload.CreatedAt.Format(time.RFC3339)
 
-	// Derive vehicle status from gear and speed. This is a synthetic field
-	// (not from Tesla telemetry) — it drives the frontend's driving/parked UI.
-	fields["status"] = deriveVehicleStatus(fields)
+	// Derive vehicle status from gear and speed only when one of those
+	// fields is present in this update. Tesla sends fields only on change,
+	// so most updates won't contain gear or speed. Injecting "parked"
+	// into those updates would override the frontend's current status and
+	// cause the driving/parked UI to flicker.
+	if _, hasGear := fields["gearPosition"]; hasGear {
+		fields["status"] = deriveVehicleStatus(fields)
+	} else if _, hasSpeed := fields["speed"]; hasSpeed {
+		fields["status"] = deriveVehicleStatus(fields)
+	}
 
 	msg, err := marshalWSMessage(msgTypeVehicleUpdate, vehicleUpdatePayload{
 		VehicleID: vehicleID,


tnando:
**Claude finished @tnando's task in 1m 7s** —— [View job](https://github.com/tnando/my-robo-taxi-telemetry/actions/runs/23677363163)

---
## Review Summary

**Verdict: APPROVE**

**Domain focus:** none (general review only)

### Findings

#### Critical (must fix)
None.

#### Warnings (should fix)
- **No test for the new conditional path.** The existing `TestBroadcaster_HandleTelemetry` sends an event with both `speed` and `gear` present, so it exercises the `hasGear` branch. However, there's no test verifying that an update **without** gear/speed correctly **omits** the `status` field. The `deriveVehicleStatus` unit tests cover the function itself, but the new guard logic in `handleTelemetry` is untested. Consider adding a subtest that publishes a telemetry event with only non-gear/non-speed fields (e.g., `soc`, `insideTemp`) and asserts `status` is absent from the broadcast payload.

#### Suggestions
- **Minor simplification** — the two branches can be collapsed since they do the same thing:
  ```go
  if _, hasGear := fields["gearPosition"]; hasGear {
      fields["status"] = deriveVehicleStatus(fields)
  } else if _, hasSpeed := fields["speed"]; hasSpeed {
      fields["status"] = deriveVehicleStatus(fields)
  }
  ```
  could become:
  ```go
  _, hasGear := fields["gearPosition"]
  _, hasSpeed := fields["speed"]
  if hasGear || hasSpeed {
      fields["status"] = deriveVehicleStatus(fields)
  }
  ```
  This is a pure style nit — not blocking.

### General Review Notes

- ✅ **File size:** `broadcaster.go` is 291 lines — under the 300-line limit
- ✅ **Error handling:** No changes to error paths; existing wrapping is correct
- ✅ **Comment quality:** The new comment clearly explains the *why* — Tesla's sparse field updates and the resulting flicker
- ✅ **No CLAUDE.md violations:** Dependencies, imports, and patterns all look correct
- ✅ **No security concerns:** No VIN logging changes, no new external inputs
- ✅ **Correct fix approach:** Addressing this at the server-side broadcast layer (rather than requiring frontend changes) is the right call — it avoids sending misleading data to all clients

Good, focused bug fix. The root cause analysis in the PR description is excellent.

PR #113, opened 2026-03-28
Fix route polyline + persist navigation fields to DB

tnando:
## Summary

Fixes three issues affecting the driving UI:

### 1. Polyline decoder tolerates truncated routes
Tesla truncates long route polylines. Instead of discarding the entire route on truncation, the decoder now returns all successfully decoded coordinate pairs. Routes now display on the map even when truncated.

### 2. Navigation fields persisted to Vehicle table
`destinationName`, `destinationLatitude/Longitude`, `originLatitude/Longitude` were broadcast via WebSocket but never written to the DB. On page refresh, all navigation data was lost. Now persisted alongside other telemetry fields.

### 3. Frontend mapper includes navigation coordinates
The vehicle mapper now reads destination/origin lat/lng from Prisma so initial page loads have navigation data.

## Test plan
- [x] All Go tests pass (polyline truncation test updated)
- [x] `golangci-lint` clean
- [ ] Drive test: route polyline appears, destination name shows


tnando:
**Claude finished @tnando's task in 1m 0s** —— [View job](https://github.com/tnando/my-robo-taxi-telemetry/actions/runs/23677929832)

---
### PR Review: Fix route polyline + persist navigation fields to DB

- [x] Check CLAUDE.md rules compliance
- [x] Review code quality and correctness
- [x] Post verdict

---

## Review Summary

**Verdict: APPROVE**

**Domain focus:** No agent labels — general review only

### Findings

#### Critical (must fix)

None.

#### Warnings (should fix)

None.

#### Suggestions

1. **`deref` helpers are unexported and undocumented** — `derefInt`, `derefFloat`, `derefString` in `queries.go` are fine as internal helpers, but a single generic `deref[T]` function (Go 1.23+ supports generics) would eliminate the three near-identical functions. Minor DRY opportunity, not blocking.

2. **Polyline: empty string now returns an error instead of `nil, nil`** — This is a behavior change (empty string previously returned `nil, nil`; now returns `nil, error`). The test was updated to reflect this, which is correct. Just flagging that callers should handle this. Looks like the test covers it well.

### General Checks

| Check | Status |
|-------|--------|
| Error handling with context | ✅ `DecodePolyline` wraps with context |
| No `any`/`interface{}` misuse | ✅ `any` only used in `updateColumn.val` where it's genuinely polymorphic (int/float64/string) |
| Context propagation | ✅ N/A for changed files (no I/O functions added) |
| Files under 300 lines | ✅ `queries.go` 161, `types.go` 126, `field_mapper.go` 121, `polyline.go` 77 |
| Dependency direction | ✅ `store` → `events`/`telemetry` types only |
| No `panic()` | ✅ |
| No gorilla/websocket | ✅ |
| Secrets/VINs | ✅ No VIN logging added |
| Table-driven tests | ✅ `polyline_test.go` uses table-driven with `t.Run` |
| SQL injection | ✅ Column names are hardcoded constants; values are parameterized via `$N` |

### Code Quality Notes

- **`buildTelemetryUpdate` refactor is a nice improvement.** The old code had repetitive `if u.Field != nil { addClause(...) }` blocks for each field. The new `updateColumns` + loop pattern is DRY and makes adding new fields trivial. Good refactor.

- **Polyline truncation tolerance is well-designed.** Breaking out of the loop on decode error and returning accumulated coords is the right approach for Tesla's truncated polylines. The "zero coords = error" guard at the end is a good safety net.

- **Scan order matches SELECT column order** in `vehicle_repo.go` — verified the new destination/origin fields are inserted in the correct position before `lastUpdated`.

- **Test schema updated** in `db_test.go` to match the new columns — integration tests will exercise the full path.

PR #129, opened 2026-03-30
Debug: log raw RouteLine encoding

tnando:
Log the raw string preview + hex bytes to determine if Tesla uses Google Encoded Polyline or a different format.

### DIFF
--- a/internal/ws/field_mapping.go
+++ b/internal/ws/field_mapping.go
@@ -1,6 +1,7 @@
 package ws
 
 import (
+	"fmt"
 	"log/slog"
 	"math"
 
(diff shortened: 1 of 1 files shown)


tnando:
**Claude finished @tnando's task in 44s** —— [View job](https://github.com/tnando/my-robo-taxi-telemetry/actions/runs/23723662736)

---
### PR Review: Debug: log raw RouteLine encoding

- [x] Review against project rules (CLAUDE.md)
- [x] Post verdict

---

## Review Summary

**Verdict: APPROVE**

**Domain focus:** none (general review only)

### Findings

#### Critical (must fix)
_None._

#### Warnings (should fix)

- **Potential VIN/sensitive data in `raw_preview`**: The `raw_preview` field logs raw telemetry string content. This is fine for a temporary debug PR, but ensure this logging is removed before merging to production. If the raw string could ever contain vehicle-identifying data, it should be redacted per CLAUDE.md rules. Since this is explicitly a debug/investigation PR to determine encoding format, this is acceptable as a short-lived change.

#### Suggestions

- **Consider using `slog.Debug` instead of `slog.Info`**: Since this is diagnostic logging to investigate encoding format, `slog.Debug` would be more appropriate — it would be silenced in production by default and wouldn't require a follow-up PR to remove the noise. The existing diagnostic log on line 147 also uses `Info`, so this is consistent with the current pattern, but both could benefit from being `Debug` level.

- **Minor: `fmt` import for a single Sprintf**: The `fmt` package was added solely for `fmt.Sprintf("%x", ...)`. An alternative is `encoding/hex.EncodeToString(...)` which is slightly more idiomatic for hex encoding, but `fmt.Sprintf` is perfectly fine for a debug log.

### Notes

This is a small, well-scoped debug PR — 11 lines added to log raw string preview and hex bytes to diagnose Tesla's RouteLine encoding format. The code is clean: it correctly bounds the preview to 100 chars and hex to 20 bytes using `min()`, uses structured logging via `slog`, and doesn't touch any business logic. No test changes needed for temporary debug logging.

The file remains well under the 300-line limit (218 lines). Imports are correctly grouped. No security, concurrency, or architectural concerns.

**Recommendation:** Approve for investigation purposes. Open a follow-up to either remove the debug logging or downgrade to `slog.Debug` once the encoding format is determined.



---

PR #135, opened 2026-03-31
Fix status derivation and clear stale nav data on Tesla invalid

tnando:
## Summary
- **Bug 1:** Remove the `else if hasSpeed` branch from status derivation in the broadcaster. When a speed=0 update arrived without a gear field, `deriveVehicleStatus` incorrectly returned "parked" because gear was empty string. Status is now only derived when `gearPosition` is present, letting the frontend keep the last known status for speed-only updates.
- **Bug 2:** Change the decoder to return `TelemetryValue{Invalid: true}` instead of `ErrInvalidValue` when Tesla marks a datum as invalid. This allows invalid fields to flow through to `mapFieldsForClient`, which now emits `nil` for nav fields (destinationName, routeLine, milesToArrival, minutesToArrival, originLocation, destinationLocation) so the frontend clears stale destination and route data when navigation is cancelled.

## Test plan
- [ ] `TestStatusDerivation_SpeedOnlyDoesNotInjectStatus` — verifies speed-only updates at 0 mph do NOT inject a status field, while gear updates still derive status correctly
- [ ] `TestMapFieldsForClient_InvalidNavFieldsClear` — verifies all six nav field types produce nil clearing values when marked invalid, and non-nav invalid fields are silently dropped
- [ ] `TestDecoder_DecodePayload_InvalidDatum` — updated to expect Invalid=true in fields map instead of a field error
- [ ] Existing `TestDeriveVehicleStatus` and `TestFieldMapping` continue to pass unchanged

### DIFF
--- a/internal/ws/broadcaster_test.go
+++ b/internal/ws/broadcaster_test.go
@@ -1121,11 +1121,6 @@ func TestStatusDerivation_SpeedOnlyDoesNotInjectStatus(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
 			clientFields := mapFieldsForClient(tt.fields)
-			_, hasStatus := clientFields["gearPosition"]
-			if tt.wantStatus && !hasStatus {
-				// Gear must be present for status to be injected.
-				// If gear is in input, gearPosition is in output.
-			}
 
 			// Simulate what handleTelemetry does.
 			if _, hasGear := clientFields["gearPosition"]; hasGear {


tnando:
**Claude finished @tnando's task in 56s** —— [View job](https://github.com/tnando/my-robo-taxi-telemetry/actions/runs/23775631877)

---
### PR Review: Fix status derivation and clear stale nav data on Tesla invalid

- [x] Check against CLAUDE.md project rules
- [x] Review code quality and test coverage
- [x] Post verdict

---

## Review Summary

**Verdict: APPROVE**

**Domain focus:** No agent labels — general review only.

### Findings

#### Critical (must fix)
None.

#### Warnings (should fix)
- **`broadcaster.go` is 302 lines** (limit is 300 per CLAUDE.md). However, this PR actually *reduced* the file from 307 → 302 lines by removing the `else if` branch, so this is a pre-existing issue that this PR improves. Not blocking.

#### Suggestions
- **`navClearFields` partially duplicates `locationFieldSplit` and `internalToClientField`:** The nav clear map repeats the same field name mappings that already exist in other maps (e.g. `"milesToArrival" → "tripDistanceRemaining"` mirrors `internalToClientField`, and `"originLocation" → ["originLatitude", "originLongitude"]` mirrors `locationFieldSplit`). This is fine for now since the semantics differ (clear vs. translate vs. split), but if more nav fields are added, consider deriving clear keys from the existing maps to avoid drift.
- **Test helper dead code (lines 1130-1133):** The `hasGear`/`hasStatus` block inside `TestStatusDerivation_SpeedOnlyDoesNotInjectStatus` doesn't do anything — the `if tt.wantStatus && !hasStatus` branch has an empty body and comments. Consider removing or replacing with an assertion.

### Code Quality Notes

**Bug fixes are well-reasoned:**
1. The status derivation fix correctly identifies that speed-only updates lack gear context, making `deriveVehicleStatus` unreliable (it would default to "parked" with empty gear string). Only deriving status when gear is present is the right call.
2. Converting `ErrInvalidValue` → `TelemetryValue{Invalid: true}` is a clean approach that lets invalid signals propagate through the pipeline without error-path short-circuiting, enabling the nav-clearing behavior downstream.

**Tests are thorough:**
- Table-driven with `t.Run` and `t.Parallel` ✅
- Tests cover both positive and negative cases (nav fields clear, non-nav fields dropped, mixed valid+invalid) ✅
- Decoder test properly updated to match new semantics ✅
- Test naming follows project conventions ✅

**Project rules compliance:**
- Error handling: correct — invalid values are no longer swallowed as errors ✅
- No `any`/`interface{}` misuse ✅
- No new globals, no `init()` ✅
- Files under 300 lines (field_mapping.go at 235, telemetry_event.go at 33) ✅
- Dependency direction correct (events → ws → events types only) ✅
- No secrets, no VIN logging issues ✅

Clean, well-tested bug fix. Approving.

PR #137, opened 2026-03-31
Clear stale nav route when RouteLine decode fails

tnando:
Fixes stale route/destination persisting after Tesla nav is cancelled. Also handles the 24-byte tag-0x12 RouteLine Tesla sends on nav clear.

### DIFF
--- a/internal/ws/broadcaster_test.go
+++ b/internal/ws/broadcaster_test.go
@@ -520,15 +520,19 @@ func TestFieldMapping(t *testing.T) {
 			},
 		},
 		{
-			name: "empty routeLine is skipped",
+			name: "empty routeLine clears navRouteCoordinates",
 			fields: map[string]events.TelemetryValue{
 				"routeLine": {StringVal: ptrString("")},
 			},
-			wantKeys: nil,
+			wantKeys: []string{"navRouteCoordinates"},
 			check: func(t *testing.T, result map[string]any) {
 				t.Helper()
-				if _, ok := result["navRouteCoordinates"]; ok {
-					t.Fatal("expected no navRouteCoordinates for empty routeLine")
+				val, ok := result["navRouteCoordinates"]
+				if !ok {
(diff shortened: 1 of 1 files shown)


tnando:
**Claude finished @tnando's task in 37s** —— [View job](https://github.com/tnando/my-robo-taxi-telemetry/actions/runs/23776501341)

---
### PR Review: #137 — Clear stale nav route when RouteLine decode fails

- [x] Read PR diff
- [x] Read full file for context
- [x] Check CLAUDE.md rules
- [x] Post review verdict

---

## Review Summary

**Verdict: APPROVE**

**Domain focus:** General review (no agent labels)

### Findings

#### Critical (must fix)
_None._

#### Warnings (should fix)
_None._

#### Suggestions

- **Doc comment slightly stale** (`internal/ws/field_mapping.go:133`): The `decodeRouteLineField` doc says _"Empty or nil strings are silently skipped"_ but empty strings now actively set `navRouteCoordinates` to `nil` — which is the correct behavior, but the comment should reflect it (e.g., _"Empty strings clear the route; nil StringVal is logged as unexpected."_). Very minor nit, not blocking.

### Notes

Clean, focused bug fix. Both changes are correct:

1. **Empty RouteLine → clear route** (line 141): When Tesla sends an empty string, the frontend now receives `navRouteCoordinates: nil` to clear the stale route. Previously this silently returned, leaving a phantom route on the map.

2. **Decode failure → clear route** (line 154): When Tesla sends an undecodable RouteLine (e.g., the 24-byte tag-0x12 payload on nav cancel), the route is now explicitly cleared instead of leaving stale data. The log message was also corrected to match the function name (`decodeRouteLineField` instead of `mapFieldsForClient`).

Both paths align with the existing `navClearFields` pattern used for `val.Invalid` fields. File stays well under the 300-line limit at 241 lines. No CLAUDE.md violations.

PR #148, opened 2026-04-03
#146 Clear stale nav fields in DB when Tesla marks them invalid

### LINKED ISSUE #146: Clear stale nav fields in DB when Tesla marks them invalid
## Problem

When Tesla cancels navigation, it marks nav fields as invalid. The WebSocket broadcaster correctly sends nil to clear the frontend, but the store Writer ignores the Invalid flag — stale values (destinationName, etaMinutes, etc.) persist in the DB Vehicle table. On page refresh, stale "Heading to Amorino Gelato" and "4 min ETA" load from the DB.

## Root Cause

`mapTelemetryToUpdate()` in `field_mapper.go` never checks `TelemetryValue.Invalid`. Invalid fields produce nil VehicleUpdate pointers, which `buildTelemetryUpdate()` skips (nil = not provided). The DB column is never touched.

## Fix

When `TelemetryValue.Invalid == true` for a nav field, the store Writer must write NULL to the corresponding DB column. Add a `ClearFields []string` slice to VehicleUpdate that lists column names to explicitly SET to NULL in the SQL update.

### Changes needed:

1. **`internal/store/types.go`** — Add `ClearFields []string` to VehicleUpdate
2. **`internal/store/field_mapper.go`** — When `val.Invalid == true` and field is a nav field, append the DB column name to `ClearFields`
3. **`internal/store/queries.go`** — In `buildTelemetryUpdate()`, add SET clauses for ClearFields: `"column" = NULL`
4. **`internal/store/writer_coalesce.go`** — Merge ClearFields in `mergeUpdate()`

### Nav fields to clear:
- destinationName
- destinationLatitude, destinationLongitude
- originLatitude, originLongitude

### Missing field mappings to add:
- etaMinutes (from minutesToArrival)
- tripDistanceRemaining (from milesToArrival) — not currently in VehicleUpdate at all

## Acceptance Criteria
- [ ] Invalid nav fields write NULL to DB
- [ ] ETAMinutes and TripDistanceRemaining added to VehicleUpdate and DB persist path
- [ ] Page refresh after nav cancel shows no stale destination/ETA
- [ ] Tests cover invalid field → NULL DB write

tnando:
## Summary

- When Tesla cancels navigation and marks telemetry fields as `Invalid=true`, the store Writer now explicitly sets the corresponding DB columns to NULL via a new `ClearFields` mechanism on `VehicleUpdate`
- Added `navFieldColumns` map in `field_mapper.go` linking internal field names to their DB column names for clearing (destinationName, etaMinutes, tripDistanceRemaining, origin/destination lat/lng)
- Added missing field appliers for `minutesToArrival` (etaMinutes) and `milesToArrival` (tripDistanceRemaining) so these values persist during normal navigation
- `mergeUpdate` now propagates `ClearFields` through coalescing so NULLs survive batching
- `buildTelemetryUpdate` appends `SET "col" = NULL` clauses for each entry in `ClearFields`

Fixes #146

## Test plan

- [x] `TestMapTelemetryToUpdate_InvalidNavFields` -- verifies each nav field type populates `ClearFields` correctly, invalid non-nav fields are ignored, and mixed valid/invalid fields work
- [x] `TestBuildTelemetryUpdate_ClearFields` -- verifies SQL contains `= NULL` clauses for clear fields, both alone and mixed with regular SET clauses
- [x] `TestBuildTelemetryUpdate_NewNavFields` -- verifies etaMinutes and tripDistanceRemaining appear in generated SQL
- [x] `TestMergeUpdate_ClearFields` -- verifies ClearFields survive coalescing (appended, not overwritten)
- [x] `TestMergeUpdate_NavFields` -- verifies all nav fields merge correctly in coalescing
- [x] New applier tests for minutesToArrival and milesToArrival in existing `TestMapTelemetryToUpdate`

### DIFF
--- a/internal/store/field_mapper.go
+++ b/internal/store/field_mapper.go
@@ -34,11 +34,11 @@ var fieldAppliers = map[telemetry.FieldName]fieldApplier{
 // names that should be SET NULL when the vehicle marks the field invalid
 // (e.g. navigation cancelled).
 var navFieldColumns = map[string][]string{
-	"destinationName":     {"destinationName"},
-	"minutesToArrival":    {"etaMinutes"},
-	"milesToArrival":      {"tripDistanceRemaining"},
-	"originLocation":      {"originLatitude", "originLongitude"},
-	"destinationLocation": {"destinationLatitude", "destinationLongitude"},
+	string(telemetry.FieldDestinationName): {"destinationName"},
+	string(telemetry.FieldMinutesToArrival): {"etaMinutes"},
+	string(telemetry.FieldMilesToArrival):   {"tripDistanceRemaining"},
+	string(telemetry.FieldOriginLocation):   {"originLatitude", "originLongitude"},
+	string(telemetry.FieldDestLocation):     {"destinationLatitude", "destinationLongitude"},
 }
 
 // mapTelemetryToUpdate converts a map of telemetry field values into a
--- a/internal/store/queries.go
+++ b/internal/store/queries.go
@@ -139,9 +139,15 @@ func buildTelemetryUpdate(vin string, u VehicleUpdate) (query string, args []any
 	var setClauses []string
 	argIdx := 1
 
+	// Build a set of columns to clear so we can skip them in the regular loop.
+	clearSet := make(map[string]bool, len(u.ClearFields))
+	for _, col := range u.ClearFields {
+		clearSet[col] = true
+	}
+
 	for _, col := range updateColumns(u) {
-		if col.val == nil {
-			continue
+		if col.val == nil || clearSet[col.col] {
+			continue // skip nil values AND columns being explicitly cleared
 		}
 		// %q produces Go double-quoted strings which match PostgreSQL's
 		// double-quoted identifier syntax. Column names are hardcoded constants.


tnando:
**Claude finished @tnando's task in 1m 38s** —— [View job](https://github.com/tnando/my-robo-taxi-telemetry/actions/runs/23935393069)

---
### PR Review: #146 Clear stale nav fields in DB when Tesla marks them invalid

- [x] Gather PR diff and context
- [x] Review code against general Go quality standards
- [x] Review against `agent:go-engineer` domain criteria
- [x] Review against `agent:testing` domain criteria
- [x] Post verdict

---

## Review Summary

**Verdict: APPROVE**

**Domain focus:** `agent:go-engineer`, `agent:testing`

### Findings

#### Critical (must fix)
None.

#### Warnings (should fix)

1. **Duplicate SET columns when coalescing value + clear for the same column** (`writer_coalesce.go:78-81`, `queries.go:155-157`)

   If update₁ sets `DestinationName = "Home"` and update₂ marks `destinationName` invalid (ClearFields), after `mergeUpdate` the merged `VehicleUpdate` will have **both** `DestinationName = &"Home"` **and** `ClearFields = ["destinationName"]`. The generated SQL will be:
   ```sql
   SET "destinationName" = $1, "destinationName" = NULL, ...
   ```
   This works today because PostgreSQL uses the **last** assignment for duplicate columns, and `ClearFields` are appended after regular columns — so NULL wins. However, this relies on an implementation detail rather than explicit intent. Consider either:
   - Clearing the corresponding pointer field(s) when appending to `ClearFields` during merge, **or**
   - Skipping columns in `updateColumns` that appear in `ClearFields` during `buildTelemetryUpdate`

   The current behavior is **correct** but **fragile**. If someone reorders the clause-building logic, the invariant breaks silently.

2. **`navFieldColumns` uses raw strings, not `telemetry.FieldName` constants** (`field_mapper.go:36-42`)

   `fieldAppliers` uses typed constants (`telemetry.FieldMinutesToArrival`), but `navFieldColumns` uses raw string keys (`"minutesToArrival"`). If a constant's string value is ever refactored, `navFieldColumns` won't get a compile error — it'll silently stop matching. Consider keying `navFieldColumns` on `telemetry.FieldName` (or at minimum, using `string(telemetry.FieldMinutesToArrival)` as keys) so the compiler catches drift.

#### Suggestions

3. **Use `slices.Sort` instead of custom bubble sort** (`field_mapper_test.go:348-356`)

   Go 1.21+ provides `slices.Sort` in the standard library, which would replace the hand-rolled `sortStrings` helper. Minor nit — test code, not blocking.

4. **Vehicle read model doesn't include new fields** (`types.go` `Vehicle` struct, `queries.go` `vehicleSelectColumns`)

   `EtaMinutes` and `TripDistRemaining` are written to the DB but the `Vehicle` read struct and `vehicleSelectColumns` don't include them. If the telemetry server ever needs to read these back (e.g., for WebSocket broadcast of current state), they'll be missing. Likely intentional/out-of-scope for this PR, but worth tracking.

5. **ClearFields could accumulate duplicates during repeated merges** (`writer_coalesce.go:79-81`)

   If multiple updates for the same VIN both mark the same nav field invalid, `ClearFields` will contain duplicates (e.g., `["etaMinutes", "etaMinutes"]`). Harmless in SQL but wasteful. A small dedup or set-based approach would keep it clean.

### Domain-Specific Notes

**Go Engineer:**
- Clean use of the existing applier pattern — `applyFloat` follows the same composition style as `applyFloatAsInt` and `applyString`. Good.
- `mergeUpdate` remains straightforward field-by-field; the ClearFields append is the right approach for coalescing.
- No goroutine changes, no new concurrency concerns.

**Testing:**
- Excellent test coverage: all new code paths (invalid nav fields, ClearFields in SQL, coalescing, new field appliers) have dedicated table-driven tests.
- Tests properly handle map iteration nondeterminism with sorting.
- The "mix of invalid and valid fields" test case verifies both sides of the logic (clear + apply) in a single scenario — good edge case coverage.
- Test naming follows `TestUnit_Method_Scenario` convention.
- `TestMergeUpdate_NavFields` isn't table-driven (single case), which is fine for a comprehensive integration-style check of all nav fields together.

### Overall

Well-structured PR. The ClearFields mechanism is a clean approach to the "Tesla cancels navigation" problem. The coalescing edge case (warning #1) is the only thing I'd recommend addressing before or shortly after merge — everything else is minor.

PR #161, opened 2026-04-10
MYR-8 Author vehicle-state-schema.md and canonical JSON Schema

tnando:
## Summary
- Fills `docs/contracts/vehicle-state-schema.md` with the canonical VehicleState schema (25 fields, 4 atomic groups, 12 consistency predicates)
- Adds `docs/contracts/schemas/vehicle-state.schema.json` (draft-2020-12) — machine-readable schema with extension keywords for classification/encryption/grouping
- Documents TS generation via `json-schema-to-typescript` and Swift Codable generation strategy

## Linear Issue
MYR-8

## Agents Involved
- [x] `sdk-architect` — schema design, atomic group boundaries, consistency predicates
- [ ] `contract-guard` — drift enforcement (will validate on PR)

## Contract Impact
- [x] Updated: `docs/contracts/vehicle-state-schema.md` (placeholder → Draft v1)
- [x] Added: `docs/contracts/schemas/vehicle-state.schema.json`

## Key architectural decisions
- **`speed` excluded from GPS atomic group** — matches current code behavior (speed broadcasts independently at 2s; GPS uses 10m delta filter). This is a documented divergence from NFR-3.1 wording.
- **`navRouteCoordinates`** uses `[lng, lat]` order (GeoJSON/Mapbox), server converts on decode from Tesla's `[lat, lng]`
- **`locationName`/`locationAddress`** are ungrouped — derived async via reverse-geocoding, can't be atomically delivered with GPS
- **`chargingState`/`timeToFull`** deferred — Tesla emits but not currently persisted; charge group is `chargeLevel + estimatedRange` only

## Test plan
- [ ] Verify `contract-guard` passes
- [ ] Cross-check field list against `internal/store/types.go`
- [ ] Validate JSON Schema syntax (draft-2020-12)
- [ ] Confirm classification anchors match data-classification.md tiers

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ LDFLAGS     := -s -w \
                -X main.commit=$(COMMIT) \
                -X main.date=$(DATE)
 
-.PHONY: build test lint vet run clean proto help
+.PHONY: build test lint vet run clean proto gen-ts-types help
 
 ## build: Compile the telemetry-server binary
 build:
@@ -37,6 +37,15 @@ run: build
 ## proto: Generate Go types from Tesla protobuf definitions
 proto:
 	./scripts/generate-proto.sh
+
+## gen-ts-types: Generate TypeScript types from vehicle-state.schema.json (for @myrobotaxi/sdk)
+gen-ts-types:
+	@command -v npx >/dev/null 2>&1 || { echo "npx not found — install Node.js"; exit 1; }
+	npx --yes json-schema-to-typescript@^15 \
+		docs/contracts/schemas/vehicle-state.schema.json \
+		--output sdk/typescript/src/types/vehicle-state.ts \
+		--style.singleQuote --style.semi --additionalProperties false
+	@echo "Generated sdk/typescript/src/types/vehicle-state.ts"
 
 ## clean: Remove build artifacts
 clean:
--- a/Makefile
+++ b/Makefile
@@ -41,10 +41,11 @@ proto:
 ## gen-ts-types: Generate TypeScript types from vehicle-state.schema.json (for @myrobotaxi/sdk)
 gen-ts-types:
 	@command -v npx >/dev/null 2>&1 || { echo "npx not found — install Node.js"; exit 1; }
-	npx --yes json-schema-to-typescript@^15 \
-		docs/contracts/schemas/vehicle-state.schema.json \
-		--output sdk/typescript/src/types/vehicle-state.ts \
-		--style.singleQuote --style.semi --additionalProperties false
+	@mkdir -p sdk/typescript/src/types
+	npx --yes --package=json-schema-to-typescript@^15 json2ts \
+		-i docs/contracts/schemas/vehicle-state.schema.json \
+		-o sdk/typescript/src/types/vehicle-state.ts \
+		--no-additionalProperties
 	@echo "Generated sdk/typescript/src/types/vehicle-state.ts"
 
 ## clean: Remove build artifacts
--- a/docs/contracts/schemas/vehicle-state.schema.json
+++ b/docs/contracts/schemas/vehicle-state.schema.json
@@ -33,21 +33,27 @@
       "examples": ["Optimus"]
     },
     "model": {
-      "type": "string",
-      "description": "Vehicle model (e.g., 'Model 3', 'Model Y').",
+      "type": ["string", "null"],
+      "description": "Vehicle model (e.g., 'Model 3', 'Model Y'). SPEC-ONLY until MYR-24: not currently loaded by the Go Vehicle struct in internal/store/types.go. Will be nullable until MYR-24 lands; consumers MUST tolerate null.",
       "x-classification": "P0",
+      "x-spec-only": true,
+      "x-spec-only-issue": "MYR-24",
       "examples": ["Model 3"]
     },
     "year": {
-      "type": "integer",
-      "description": "Model year.",
+      "type": ["integer", "null"],
+      "description": "Model year. SPEC-ONLY until MYR-24: not currently loaded by the Go Vehicle struct in internal/store/types.go. Will be nullable until MYR-24 lands; consumers MUST tolerate null.",
       "x-classification": "P0",
+      "x-spec-only": true,
+      "x-spec-only-issue": "MYR-24",
       "examples": [2024]
     },
     "color": {
-      "type": "string",
-      "description": "Vehicle exterior color.",
+      "type": ["string", "null"],
+      "description": "Vehicle exterior color. SPEC-ONLY until MYR-24: not currently loaded by the Go Vehicle struct in internal/store/types.go. Will be nullable until MYR-24 lands; consumers MUST tolerate null.",
       "x-classification": "P0",
+      "x-spec-only": true,
+      "x-spec-only-issue": "MYR-24",
       "examples": ["Midnight Silver Metallic"]
     },
     "status": {
@@ -75,7 +81,7 @@
     },
     "latitude": {
       "type": "number",
-      "description": "Current GPS latitude. Encrypted at rest (AES-256-GCM). Nullable on initial connect before first GPS fix.",
+      "description": "Current GPS latitude. Encrypted at rest (AES-256-GCM). Non-nullable: on initial connect before the first GPS fix, latitude/longitude default to 0. Per the '0,0 = no fix' convention (§2.3), SDK consumers MUST treat latitude == 0 && longitude == 0 as 'no position available' rather than a valid location in the Gulf of Guinea.",
       "x-unit": "degrees",
       "x-classification": "P1",
       "x-encrypted": true,
@@ -85,7 +91,7 @@
     },
     "longitude": {
       "type": "number",
-      "description": "Current GPS longitude. Encrypted at rest (AES-256-GCM). Nullable on initial connect before first GPS fix.",
+      "description": "Current GPS longitude. Encrypted at rest (AES-256-GCM). Non-nullable: on initial connect before the first GPS fix, latitude/longitude default to 0. Per the '0,0 = no fix' convention (§2.3), SDK consumers MUST treat latitude == 0 && longitude == 0 as 'no position available' rather than a valid location in the Gulf of Guinea.",
       "x-unit": "degrees",
       "x-classification": "P1",
       "x-encrypted": true,
@@ -94,14 +100,18 @@
       "x-atomic-group": "gps"
     },
     "locationName": {
-      "type": "string",
-      "description": "Reverse-geocoded place name for current location (e.g., 'Home', 'Whole Foods Market'). Derived from GPS coordinates server-side.",
-      "x-classification": "P1"
+      "type": ["string", "null"],
+      "description": "Reverse-geocoded place name for current location (e.g., 'Home', 'Whole Foods Market'). Derived from GPS coordinates server-side. SPEC-ONLY until MYR-24: not currently loaded by the Go Vehicle struct in internal/store/types.go. Will be nullable until MYR-24 lands; consumers MUST tolerate null.",
+      "x-classification": "P1",
+      "x-spec-only": true,
+      "x-spec-only-issue": "MYR-24"
     },
     "locationAddress": {
-      "type": "string",
-      "description": "Reverse-geocoded street address for current location. Derived from GPS coordinates server-side.",
-      "x-classification": "P1"
+      "type": ["string", "null"],
+      "description": "Reverse-geocoded street address for current location. Derived from GPS coordinates server-side. SPEC-ONLY until MYR-24: not currently loaded by the Go Vehicle struct in internal/store/types.go. Will be nullable until MYR-24 lands; consumers MUST tolerate null.",
+      "x-classification": "P1",
+      "x-spec-only": true,
+      "x-spec-only-issue": "MYR-24"
     },
     "gearPosition": {
       "type": ["string", "null"],
@@ -147,10 +157,12 @@
       "minimum": 0
     },
     "fsdMilesToday": {
-      "type": "number",
-      "description": "Full Self-Driving miles driven since last counter reset (typically daily).",
+      "type": ["number", "null"],
+      "description": "Full Self-Driving miles driven since last counter reset (typically daily). SPEC-ONLY until MYR-24: not currently loaded by the Go Vehicle struct in internal/store/types.go. Will be nullable until MYR-24 lands; consumers MUST tolerate null.",
       "x-unit": "miles",
       "x-classification": "P0",
+      "x-spec-only": true,
+      "x-spec-only-issue": "MYR-24",
       "minimum": 0
     },
     "destinationName": {
@@ -162,9 +174,11 @@
     },
     "destinationAddress": {
       "type": ["string", "null"],
-      "description": "Street address of the active navigation destination. Null when no navigation is active.",
+      "description": "Street address of the active navigation destination. Null when no navigation is active. SPEC-ONLY until MYR-24: not currently loaded by the Go Vehicle struct in internal/store/types.go; will always be null in snapshots until MYR-24 lands.",
       "x-classification": "P1",
-      "x-atomic-group": "navigation"
+      "x-atomic-group": "navigation",
+      "x-spec-only": true,
+      "x-spec-only-issue": "MYR-24"
     },
     "destinationLatitude": {
       "type": ["number", "null"],
@@ -276,7 +290,7 @@
       ],
       "consistency-predicates": [
         "chargeLevel and estimatedRange MUST both be present in the DB snapshot (NFR-3.5). They are non-nullable with default 0.",
-        "If chargeLevel is 0 and estimatedRange is 0, this represents either a genuinely depleted battery or an initial-state default -- the UI SHOULD treat this as valid data."
+        "chargeLevel: 0 is context-dependent. With status == 'offline', the UI SHOULD interpret chargeLevel: 0 as 'unknown' (no telemetry received yet, DB default). With status == 'driving' or 'parked' (i.e., the vehicle is online), chargeLevel: 0 SHOULD be interpreted as 'critical' or 'empty battery'. See §2.2 of vehicle-state-schema.md."
       ],
       "nullability": "Non-nullable. Default 0 on initial vehicle creation."
     },
@@ -306,68 +320,6 @@
         "status is always present; gearPosition is nullable (null = vehicle asleep or gear not yet reported)."
       ],
       "nullability": "gearPosition is nullable. status is non-nullable (default 'offline')."
-    }
-  },
-  "$defs": {
-    "NavigationGroup": {
-      "type": "object",
-      "description": "Sub-schema for the navigation atomic group. Used for validation of partial updates.",
-      "properties": {
-        "destinationName": { "$ref": "#/properties/destinationName" },
-        "destinationAddress": { "$ref": "#/properties/destinationAddress" },
-        "destinationLatitude": { "$ref": "#/properties/destinationLatitude" },
-        "destinationLongitude": { "$ref": "#/properties/destinationLongitude" },
-        "originLatitude": { "$ref": "#/properties/originLatitude" },
-        "originLongitude": { "$ref": "#/properties/originLongitude" },
-        "etaMinutes": { "$ref": "#/properties/etaMinutes" },
-        "tripDistanceRemaining": { "$ref": "#/properties/tripDistanceRemaining" },
-        "navRouteCoordinates": { "$ref": "#/properties/navRouteCoordinates" }
-      },
-      "dependentRequired": {
-        "destinationLatitude": ["destinationLongitude"],
-        "destinationLongitude": ["destinationLatitude"],
-        "originLatitude": ["originLongitude"],
-        "originLongitude": ["originLatitude"]
-      }
-    },
-    "ChargeGroup": {
-      "type": "object",
-      "description": "Sub-schema for the charge atomic group.",
-      "properties": {
-        "chargeLevel": { "$ref": "#/properties/chargeLevel" },
-        "estimatedRange": { "$ref": "#/properties/estimatedRange" }
-      },
-      "required": ["chargeLevel", "estimatedRange"]
-    },
-    "GpsGroup": {
-      "type": "object",
-      "description": "Sub-schema for the GPS atomic group.",
-      "properties": {
-        "latitude": { "$ref": "#/properties/latitude" },
-        "longitude": { "$ref": "#/properties/longitude" },
-        "heading": { "$ref": "#/properties/heading" }
-      },
-      "required": ["latitude", "longitude", "heading"],
-      "dependentRequired": {
-        "latitude": ["longitude"],
-        "longitude": ["latitude"]
-      }
-    },
-    "GearGroup": {
-      "type": "object",
-      "description": "Sub-schema for the gear atomic group.",
-      "properties": {
-        "gearPosition": { "$ref": "#/properties/gearPosition" },
-        "status": { "$ref": "#/properties/status" }
-      },
-      "required": ["status"]
-    },
-    "Coordinate": {
-      "type": "array",
-      "items": { "type": "number" },
-      "minItems": 2,
-      "maxItems": 2,
-      "description": "A [longitude, latitude] coordinate pair in GeoJSON/Mapbox order."
     }
   },
   "dependentRequired": {
--- a/docs/contracts/schemas/vehicle-state.schema.json
+++ b/docs/contracts/schemas/vehicle-state.schema.json
@@ -1,0 +1,379 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://myrobotaxi.com/schemas/vehicle-state.schema.json",
+  "title": "VehicleState",
+  "description": "Canonical vehicle state as consumed by MyRoboTaxi SDKs. Defines every field name, type, unit, nullability, atomic group membership, and data classification. Both the WebSocket vehicle_update messages and the REST snapshot endpoint return subsets of this schema.",
+  "type": "object",
+  "required": [
+    "vehicleId",
+    "name",
+    "status",
+    "chargeLevel",
+    "estimatedRange",
+    "speed",
+    "heading",
+    "latitude",
+    "longitude",
+    "interiorTemp",
+    "exteriorTemp",
+    "odometerMiles",
+    "lastUpdated"
+  ],
+  "properties": {
+    "vehicleId": {
+      "type": "string",
+      "description": "Opaque database identifier (cuid). NOT the VIN. All SDK API calls use this ID (FR-4.2).",
+      "x-classification": "P0",
+      "examples": ["clxyz1234567890abcdef"]
+    },
+    "name": {
+      "type": "string",
+      "description": "User-assigned vehicle name.",
+      "x-classification": "P0",
+      "examples": ["Optimus"]
+    },
+    "model": {
+      "type": "string",
+      "description": "Vehicle model (e.g., 'Model 3', 'Model Y').",
+      "x-classification": "P0",
+      "examples": ["Model 3"]
+    },
+    "year": {
+      "type": "integer",
+      "description": "Model year.",
+      "x-classification": "P0",
+      "examples": [2024]
+    },
+    "color": {
+      "type": "string",
+      "description": "Vehicle exterior color.",
+      "x-classification": "P0",
+      "examples": ["Midnight Silver Metallic"]
+    },
+    "status": {
+      "type": "string",
+      "enum": ["driving", "parked", "charging", "offline", "in_service"],
+      "description": "Derived vehicle status. Part of the gear atomic group -- updated atomically with gearPosition. When gearPosition is D or R, status is 'driving'; otherwise 'parked'. The 'charging', 'offline', and 'in_service' states are set by server-side logic.",
+      "x-classification": "P0",
+      "x-atomic-group": "gear"
+    },
+    "speed": {
+      "type": "integer",
+      "description": "Current vehicle speed. Rounded to nearest integer from Tesla's float emission.",
+      "x-unit": "mph",
+      "x-classification": "P0",
+      "minimum": 0
+    },
+    "heading": {
+      "type": "integer",
+      "description": "Compass heading. 0 = North, 90 = East, 180 = South, 270 = West.",
+      "x-unit": "degrees",
+      "x-classification": "P0",
+      "minimum": 0,
+      "maximum": 359,
+      "x-atomic-group": "gps"
+    },
+    "latitude": {
+      "type": "number",
+      "description": "Current GPS latitude. Encrypted at rest (AES-256-GCM). Nullable on initial connect before first GPS fix.",
+      "x-unit": "degrees",
+      "x-classification": "P1",
+      "x-encrypted": true,
+      "minimum": -90,
+      "maximum": 90,
+      "x-atomic-group": "gps"
+    },
+    "longitude": {
+      "type": "number",
+      "description": "Current GPS longitude. Encrypted at rest (AES-256-GCM). Nullable on initial connect before first GPS fix.",
+      "x-unit": "degrees",
+      "x-classification": "P1",
+      "x-encrypted": true,
+      "minimum": -180,
+      "maximum": 180,
+      "x-atomic-group": "gps"
+    },
+    "locationName": {
+      "type": "string",
+      "description": "Reverse-geocoded place name for current location (e.g., 'Home', 'Whole Foods Market'). Derived from GPS coordinates server-side.",
+      "x-classification": "P1"
+    },
+    "locationAddress": {
+      "type": "string",
+      "description": "Reverse-geocoded street address for current location. Derived from GPS coordinates server-side.",
+      "x-classification": "P1"
+    },
+    "gearPosition": {
+      "type": ["string", "null"],
+      "enum": ["P", "D", "R", "N", null],
+      "description": "Transmission gear position. Null when vehicle is asleep or gear has not yet been reported. When gearPosition changes, status is re-derived and both are delivered together.",
+      "x-classification": "P0",
+      "x-atomic-group": "gear"
+    },
+    "chargeLevel": {
+      "type": "integer",
+      "description": "Battery state of charge. Rounded to nearest integer from Tesla's float emission. Maps from Tesla SOC or BatteryLevel fields.",
+      "x-unit": "percent",
+      "x-classification": "P0",
+      "minimum": 0,
+      "maximum": 100,
+      "x-atomic-group": "charge"
+    },
+    "estimatedRange": {
+      "type": "integer",
+      "description": "Estimated remaining driving range based on current charge. Rounded to nearest integer.",
+      "x-unit": "miles",
+      "x-classification": "P0",
+      "minimum": 0,
+      "x-atomic-group": "charge"
+    },
+    "interiorTemp": {
+      "type": "integer",
+      "description": "Cabin interior temperature. Rounded to nearest integer.",
+      "x-unit": "fahrenheit",
+      "x-classification": "P0"
+    },
+    "exteriorTemp": {
+      "type": "integer",
+      "description": "Ambient exterior temperature. Rounded to nearest integer.",
+      "x-unit": "fahrenheit",
+      "x-classification": "P0"
+    },
+    "odometerMiles": {
+      "type": "integer",
+      "description": "Total odometer reading. Rounded to nearest integer.",
+      "x-unit": "miles",
+      "x-classification": "P0",
+      "minimum": 0
+    },
+    "fsdMilesToday": {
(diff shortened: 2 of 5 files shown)


tnando:
**Claude finished @tnando's task in 3m 4s** —— [View job](https://github.com/tnando/my-robo-taxi-telemetry/actions/runs/24225913243)

---
## Review Summary

**Verdict: REQUEST_CHANGES**

**Domain focus:** `agent:sdk-architect`

This is a high-quality contract draft — well-anchored to FRs/NFRs, thoughtful atomic-group design, and honest about divergences (speed vs. NFR-3.1, deferred `chargingState`/`tripStartTime`). The markdown narrative is excellent. However, the **JSON Schema file has several concrete inconsistencies with the markdown spec that will propagate to generated TS/Swift types**. Because this file is declared the single source of truth for code generation, those mismatches are blockers.

### Findings

#### Critical (must fix)

1. **`required` array is missing non-nullable fields from §1.1** — [`vehicle-state.schema.json` L7–L21](docs/contracts/schemas/vehicle-state.schema.json#L7-L21)
   The markdown table lists `model`, `year`, `color`, `fsdMilesToday`, `locationName`, `locationAddress` as **Nullable: No**, but none of them appear in the schema's root `required` array. Either:
   - (a) add them to `required` so generators emit non-optional fields, or
   - (b) mark them as nullable in the markdown if they can be absent on initial connect / partial updates.
   Since §1.1 says "Both the WebSocket protocol and the REST snapshot endpoint return subsets of this schema," the schema should represent the **full canonical shape** and (a) is the right fix.

2. **Consistency predicates in §3.1 are not enforceable by the current schema** — [`vehicle-state-schema.md` L156–L163](docs/contracts/vehicle-state-schema.md#L156-L163)
   `dependentRequired` only enforces *presence* of a key, not that the value is non-null. The critical predicate *"if `destinationName` is non-null, then `destinationLatitude`, `destinationLongitude`, and `navRouteCoordinates` MUST also be non-null"* is not expressible without `if/then` (or `allOf` with `not: { const: null }`). Right now the schema will happily validate `{ destinationName: "Foo", destinationLatitude: null, ... }`. Either:
   - add `if/then` clauses for each non-null predicate, or
   - explicitly annotate (in schema + markdown) that these are **semantic invariants enforced by `contract-tester`** rather than by the schema, so the contract-guard doesn't give a false sense of coverage.

3. **`$defs` sub-schemas are unreachable dead code** — [`vehicle-state.schema.json` L311–L371](docs/contracts/schemas/vehicle-state.schema.json#L311-L371)
   `NavigationGroup`, `ChargeGroup`, `GpsGroup`, `GearGroup`, and `Coordinate` are defined but never `$ref`'d anywhere in the root object. `json-schema-to-typescript` will either emit them as orphan types or drop them, creating drift between runtime validation and generated types. Either `$ref` them from the root (e.g., `allOf: [{$ref: "#/$defs/NavigationGroup"}, ...]`) or delete them until they have a consumer.

4. **Schema vs. `internal/store/types.go` gap** — [`internal/store/types.go` L25–L50](internal/store/types.go#L25-L50)
   The `Vehicle` read model does **not** contain `model`, `year`, `color`, `locationName`, `locationAddress`, `destinationAddress`, or `fsdMilesToday`. Those columns exist in the DB per `data-classification.md` §1.3, but the telemetry server's Vehicle struct never loads them, so they cannot appear in any broadcast or REST snapshot today. Options:
   - (a) File a follow-up MYR issue to extend `Vehicle` + `vehicle_repo.go` to read these columns (recommended — add it to §7.2 "Open questions").
   - (b) Carve the schema into a "telemetry-scoped" core and a "catalog-scoped" extension.
   Either way, this gap deserves explicit mention in the change log so consumers don't generate types that the server can't populate.

#### Warnings (should fix)

5. **`latitude`/`longitude` description contradicts their type** — [`vehicle-state.schema.json` L76–L95](docs/contracts/schemas/vehicle-state.schema.json#L76-L95)
   `"type": "number"` and they're in `required`, but the description says *"Nullable on initial connect before first GPS fix"*. The markdown (§1.1, §2.3) explicitly says **non-nullable**. Drop the "Nullable on initial connect" phrase from the description and point readers to the "`0,0` means no fix" convention instead.

6. **`chargeLevel: 0` semantics disagree between MD and schema** —
   - `vehicle-state-schema.md` §2.2 (L126): *"the UI should interpret `chargeLevel: 0` with `status: offline` as 'unknown'"*
   - `vehicle-state.schema.json` L279: *"the UI SHOULD treat this as valid data."*
   Pick one rule and state it in both places — ideally the §2.2 framing (context-dependent), since the schema is the source of truth for generators and the MD is authoritative for semantics.

7. **`gearPosition` enum encoding** — [`vehicle-state.schema.json` L106–L112](docs/contracts/schemas/vehicle-state.schema.json#L106-L112)
   `"type": ["string", "null"], "enum": ["P", "D", "R", "N", null]` is valid in draft-2020-12, but some older TS/Swift generators trip on mixed-type enums. If `json-schema-to-typescript` is pinned, verify it produces `"P" | "D" | "R" | "N" | null` and not `string | null`.

#### Suggestions

- **Redundant `x-atomic-group` + `x-atomic-groups`**: per-field `x-atomic-group` labels and the root `x-atomic-groups` object both encode the same membership. Fine to keep for ergonomics, but pick one as authoritative and have contract-guard verify the other is derived from it.
- **Drop `$id` placeholder if not real**: `"https://myrobotaxi.com/schemas/vehicle-state.schema.json"` should either resolve or be an `urn:` to avoid breaking future bundlers that try to fetch external refs.
- **Link §4 source-of-truth rows to the Tesla field catalog**: a cross-link to the protobuf field definitions or `field_mapper.go` would help reviewers trace every row back to real code.
- **§6.1 generation command**: consider adding a `Makefile` target (`make gen-ts-types`) so the command in the doc stays in sync with what CI actually runs.

### Domain-Specific Notes (sdk-architect)

The architectural thinking here is strong — I especially appreciate:
- The explicit `speed` divergence from NFR-3.1 with rationale (flood-vs-delay tradeoff) documented in §7.1 rather than hidden.
- Nav cancellation semantics tied to FR-2.3 with all-or-nothing clear in §3.1(3). This is exactly the right invariant.
- Per-field `x-encrypted: true` driving AES-256-GCM boundaries — this will let contract-guard mechanically verify NFR-3.23 encryption coverage.
- ISO 8601 `lastUpdated` with `format: date-time` (not Unix epoch) — consistent with the frontend-integration expectations in CLAUDE.md.

Main architectural ask before merge: **reconcile the schema with `internal/store/types.go`**. Right now an sdk-typescript agent could generate a `VehicleState` type containing `color`, `model`, `year`, `fsdMilesToday`, `locationName`, `locationAddress`, and `destinationAddress` that the Go server literally cannot populate — that's the exact contract drift this file is meant to prevent. A MYR issue to extend the Go read model, linked from §7.2, would unblock the contract while making the gap explicit.

---
 · [Branch: thomasnandola/myr-8-author-vehicle-state-schemamd-canonical-vehiclestate-json](https://github.com/tnando/my-robo-taxi-telemetry/tree/thomasnandola/myr-8-author-vehicle-state-schemamd-canonical-vehiclestate-json)

tnando:
**Claude finished @tnando's task in 2m 2s** —— [View job](https://github.com/tnando/my-robo-taxi-telemetry/actions/runs/24226132399)

---
### Review Summary

**Verdict: REQUEST_CHANGES**

**Domain focus:** `agent:sdk-architect`, `agent:go-engineer`

The new commit in this push (`5271584`) is an unrelated (and well-reasoned) fix to `TestBroadcaster_DriveEndedClearsAccumulator` — that part looks good. However, **none of the six critical/warning findings from the previous review have been addressed** in either commit. The schema and markdown still diverge, and the gap against `internal/store/types.go` is still present. Re-requesting changes on the same blockers.

---

### Findings

#### Critical (must fix — still open from previous review)

1. **`required` array still missing non-nullable fields** — [`vehicle-state.schema.json` L7–L21](docs/contracts/schemas/vehicle-state.schema.json#L7-L21)
   The markdown (§1.1, L36–L41, L85–L91) lists `model`, `year`, `color`, `fsdMilesToday`, `locationName`, `locationAddress` as **Nullable: No**, but they are still absent from the schema's root `required`. Consumers generating TS/Swift types from this schema will emit these as optional, contradicting the MD. Either add them to `required` or mark them nullable in the MD — pick one and make it consistent. [Fix this →](https://claude.ai/code?q=In%20docs%2Fcontracts%2Fschemas%2Fvehicle-state.schema.json%20add%20model%2C%20year%2C%20color%2C%20fsdMilesToday%2C%20locationName%2C%20locationAddress%20to%20the%20root%20required%20array%20on%20branch%20thomasnandola%2Fmyr-8-author-vehicle-state-schemamd-canonical-vehiclestate-json&repo=tnando/my-robo-taxi-telemetry)

2. **Schema vs. `internal/store/types.go` gap — the contract-drift hazard this doc is meant to prevent** — [`internal/store/types.go` L25–L50](internal/store/types.go#L25-L50)
   Confirmed unchanged: `Vehicle` has no `Model`, `Year`, `Color`, `FsdMilesToday`, `LocationName`, `LocationAddress`, or `DestinationAddress` columns loaded. An `sdk-typescript` agent generating types from this schema will produce fields the Go server literally cannot populate today. `VehicleUpdate` (L55–L78) does carry `LocationName`/`LocationAddr`, so the write path exists, but the read model never loads them back — meaning the REST snapshot will omit them and any WebSocket replay will drop them.
   Minimum acceptable fix: file a follow-up MYR issue to extend `Vehicle` + `vehicle_repo.go`, link it from §7.2 "Open questions", and note in the change log that these fields are "spec-only until MYR-XXX lands." Otherwise contract-guard is giving a false-green. [Fix this →](https://claude.ai/code?q=In%20docs%2Fcontracts%2Fvehicle-state-schema.md%20section%207.2%20document%20the%20gap%20between%20the%20schema%20and%20internal%2Fstore%2Ftypes.go%20Vehicle%20struct%20which%20is%20missing%20Model%20Year%20Color%20FsdMilesToday%20LocationName%20LocationAddress%20DestinationAddress%20on%20branch%20thomasnandola%2Fmyr-8-author-vehicle-state-schemamd-canonical-vehiclestate-json&repo=tnando/my-robo-taxi-telemetry)

3. **Consistency predicates in §3.1 still not enforceable by the schema** — [`vehicle-state.schema.json` L373–L378](docs/contracts/schemas/vehicle-state.schema.json#L373-L378)
   Only `dependentRequired` exists at the root, which enforces *key presence*, not non-null values. The critical predicate *"if `destinationName` is non-null, then `destinationLatitude/Longitude/navRouteCoordinates` MUST be non-null"* still passes for `{ destinationName: "Foo", destinationLatitude: null, ... }`. Either add `if/then/else` clauses or explicitly annotate in both files that these are **semantic invariants enforced by `contract-tester`**, not by the schema. Right now §3.1 implies schema-level enforcement that doesn't exist.

4. **`$defs` sub-schemas are still unreachable** — [`vehicle-state.schema.json` L311–L371](docs/contracts/schemas/vehicle-state.schema.json#L311-L371)
   `NavigationGroup`, `ChargeGroup`, `GpsGroup`, `GearGroup`, and `Coordinate` are defined but never `$ref`'d from the root. `json-schema-to-typescript` may emit orphan types or drop them, creating silent drift with the generated TypeScript. Either compose them into the root (e.g., via `allOf`) or delete them until a concrete consumer exists. Leaving dead schema nodes in a source-of-truth file is a maintainability trap.

#### Warnings (must fix — also still open)

5. **`latitude`/`longitude` description still contradicts the type** — [`vehicle-state.schema.json` L78, L88](docs/contracts/schemas/vehicle-state.schema.json#L78)
   Type is `"number"` (non-nullable) and both are in `required`, but descriptions still read *"Nullable on initial connect before first GPS fix."* The MD's "`0,0` means no fix" convention (§2.3) is the correct framing — update the schema descriptions to match. As written, a reader of the schema would reasonably conclude the field can be `null`, and a strict validator will reject `null` — contradiction.

6. **`chargeLevel: 0` semantics still disagree** —
   - [`vehicle-state-schema.md` L126](docs/contracts/vehicle-state-schema.md#L126): *"the UI should interpret `chargeLevel: 0` with `status: offline` as 'unknown'"*
   - [`vehicle-state.schema.json` L279](docs/contracts/schemas/vehicle-state.schema.json#L279): *"the UI SHOULD treat this as valid data."*
   Still contradictory. Pick the MD framing (context-dependent) and rewrite the schema annotation to match.

- Still worth: pick one of `x-atomic-group` vs `x-atomic-groups` as authoritative and derive the other, to prevent silent drift.
- Still worth: replace the `$id` placeholder (`https://myrobotaxi.com/schemas/...`) with a `urn:` form if it doesn't resolve.
- §6.1 should land as a `Makefile` target (`make gen-ts-types`) so CI and docs can't diverge.

---

### `broadcaster_test.go` fix (commit `5271584`) — looks good

The flake analysis is correct: `callLog` was already populated by route-point processing before `DriveEnded` was published, so `len(callLog) > 0` returned immediately and raced `handleDriveEnded`'s accumulator clear. Capturing `callLogBefore` under the read lock and waiting for strict growth is the right fix. No concerns here.

One minor suggestion (non-blocking): consider adding a short comment block above `callLogBefore` noting *why* the classic pattern races — the inline comment is clear, but a future reader tweaking the test might miss it. What you have is acceptable.

---

### Domain-Specific Notes (sdk-architect)

The architectural bones of this contract remain strong — the `speed`-vs-NFR-3.1 divergence rationale, nav cancellation atomicity, per-field encryption metadata, and ISO 8601 `lastUpdated` are all excellent. But the value of a canonical schema file is **zero if it silently diverges from the Go read model and from its own prose**. The previous review flagged six concrete mismatches; the new commit on this push addresses a test flake instead. Please push a second commit that actually touches the schema + markdown.

**Minimum bar to unblock:**
1. Reconcile `required` ↔ MD nullability (Critical #1).
2. Either extend `Vehicle` or document the gap + link a follow-up MYR (Critical #2).
3. Fix or demote the `§3.1` non-null predicates claim (Critical #3).
4. Either wire up `$defs` or delete them (Critical #4).
5. Fix the `latitude`/`longitude` and `chargeLevel: 0` text contradictions (Warnings #5, #6).

`contract-guard` should remain red until all six are addressed.

---
 · [Branch: thomasnandola/myr-8-author-vehicle-state-schemamd-canonical-vehiclestate-json](https://github.com/tnando/my-robo-taxi-telemetry/tree/thomasnandola/myr-8-author-vehicle-state-schemamd-canonical-vehiclestate-json)

tnando:
**Claude finished @tnando's task in 1m 56s** —— [View job](https://github.com/tnando/my-robo-taxi-telemetry/actions/runs/24226375201)

---
### Review Summary

**Verdict: APPROVE**

**Domain focus:** `agent:sdk-architect`, `agent:go-engineer`

All six critical/warning findings from the previous two review rounds have been resolved in commit `d4ea7de`. The schema and markdown are now internally consistent, the gap against `internal/store/types.go` is explicitly documented and tracked under MYR-24, and the §3.1 enforcement boundary is honest about what JSON Schema can and cannot express.

---

### Verification of prior findings

#### Critical (all resolved)

1. **`required` array vs nullable fields** — [`vehicle-state.schema.json` L7–L21](docs/contracts/schemas/vehicle-state.schema.json#L7-L21) ✅
   Fixed via option (b): `model`, `year`, `color`, `fsdMilesToday`, `locationName`, `locationAddress`, `destinationAddress` are all now `"type": ["<t>", "null"]` in the schema and **Yes (Spec-only, MYR-24)** in the MD (§1.1, L40–L42, L91–L93). Consistent on both sides.

2. **Schema vs `internal/store/types.go` gap** — [`internal/store/types.go` L25–L50](internal/store/types.go#L25-L50) ✅
   Confirmed unchanged (the Go struct still lacks `Model`, `Year`, `Color`, `FsdMilesToday`, `LocationName`, `LocationAddress`, `DestinationAddress`). But the gap is now explicit:
   - §1.1 SPEC-ONLY CALLOUT ([L32](docs/contracts/vehicle-state-schema.md#L32)) names all 7 fields and links MYR-24.
   - §7.2 Open questions row ([L345](docs/contracts/vehicle-state-schema.md#L345)) tracks the follow-up.
   - Each spec-only field carries `"x-spec-only": true` + `"x-spec-only-issue": "MYR-24"` in the schema, so `contract-guard` can mechanically detect the gap.
   This is the minimum bar I asked for and it's cleanly done.

3. **§3.1 non-null predicates not expressible in JSON Schema** — [`vehicle-state-schema.md` L156](docs/contracts/vehicle-state-schema.md#L156) ✅
   §3.1 now opens with an **Enforcement boundary** paragraph making it explicit that these are semantic invariants enforced by `contract-tester` at runtime and by the server's persistence layer at write-time, **NOT** by the JSON Schema. Each predicate in §3.1 is individually tagged "Semantic invariant only; not schema-enforceable." `dependentRequired` is kept in the schema ([L325–L330](docs/contracts/schemas/vehicle-state.schema.json#L325-L330)) as a "weaker best-effort hint" and the MD describes it that way. This is the correct framing.

4. **Unreachable `$defs`** — [`vehicle-state.schema.json` L311–L371](docs/contracts/schemas/vehicle-state.schema.json) ✅
   `$defs` is gone entirely. §5.2 ([L240–L242](docs/contracts/vehicle-state-schema.md#L240-L242)) now documents the deliberate absence and explains that atomic-group membership is encoded via per-field `x-atomic-group` + the root `x-atomic-groups` object. No orphan types to trip `json-schema-to-typescript`.

#### Warnings (all resolved)

5. **`latitude`/`longitude` description** — [`vehicle-state.schema.json` L84, L94](docs/contracts/schemas/vehicle-state.schema.json#L84) ✅
   Both descriptions now open with *"Non-nullable: on initial connect before the first GPS fix, latitude/longitude default to 0. Per the '0,0 = no fix' convention (§2.3), SDK consumers MUST treat latitude == 0 && longitude == 0 as 'no position available'..."* Matches §2.3 ([L138](docs/contracts/vehicle-state-schema.md#L138)) exactly.

6. **`chargeLevel: 0` semantics** — [`vehicle-state.schema.json` L293](docs/contracts/schemas/vehicle-state.schema.json#L293) ✅
   Schema's `chargeLevel` predicate now reads *"chargeLevel: 0 is context-dependent. With status == 'offline', the UI SHOULD interpret chargeLevel: 0 as 'unknown'... With status == 'driving' or 'parked' ... 'critical'..."* which matches §2.2 ([L128](docs/contracts/vehicle-state-schema.md#L128)) verbatim.

---

### `broadcaster_test.go` fix — still good

Commit `5271584`: the `callLogBefore` capture under the read lock + strict-growth wait is the correct fix for the `handleDriveEnded` accumulator clear race. The added comment block explaining *why* the classic `len(callLog) > 0` pattern races is exactly what a future maintainer will need. Note: the pre-existing `time.Sleep(100 * time.Millisecond)` at [broadcaster_test.go:1051](internal/ws/broadcaster_test.go#L1051) is out of scope for this PR but is the kind of thing CLAUDE.md "no sleep in tests" flags — worth cleaning up in a follow-up.

---

### Domain-Specific Notes (sdk-architect)

The architectural contract is now in excellent shape:
- **Honest divergence documentation**: `speed` outside the GPS group (§7.1) and the `tripStartTime` deferral are called out in their own table rows with rationale.
- **Mechanical contract-guard hooks**: `x-classification`, `x-encrypted`, `x-atomic-group`, and `x-spec-only` give `contract-guard` everything it needs to detect drift without reasoning about prose.
- **Runtime vs schema split**: the §3.1 enforcement-boundary paragraph is the right framing — semantic invariants belong in `contract-tester`, not in draft-2020-12 keywords.

### Suggestions (non-blocking)

- **`destinationAddress` inside the navigation atomic group**: it's listed in `x-atomic-groups.navigation.fields` ([L267](docs/contracts/schemas/vehicle-state.schema.json#L267)) but also marked `x-spec-only: true`. Predicate #3 in §3.1 correctly excludes it from the active-nav completeness predicate (only `destinationLatitude/Longitude/navRouteCoordinates` are required non-null), but predicate #3 ("all-or-nothing clear") as literally written would flag any active-navigation state where `destinationAddress` is null — which will be **every** state until MYR-24 lands. Consider adding a sentence to §3.1(3) saying "spec-only fields are exempt from the all-or-nothing clear invariant until their backing Go struct support lands." Otherwise the `contract-tester` author will either write a broken test or have to re-derive this logic.
- **§6.1 generation command → Makefile target**: still worth landing as `make gen-ts-types` in a follow-up so the doc and CI can't drift.
- **Pre-existing `time.Sleep(100 * time.Millisecond)` in `TestBroadcaster_DriveEndedClearsAccumulator`**: out of scope for this PR but should be replaced with a polling wait-for-condition per CLAUDE.md's no-sleep-in-tests rule.

---

### What's still required before merge

Per CLAUDE.md's Merge Policy, make sure:
- `contract-guard` passes against the updated schema + MD
- `sdk-architect` final PR review is recorded
- MYR-24 is actually filed in Linear and the `[MYR-24]` links in this PR resolve (the MD points to `https://linear.app/myrobotaxi/issue/MYR-24` — confirm that URL resolves, otherwise the "tracked follow-up" is vapor)

With those in place, this is good to merge.



---

 · [Branch: thomasnandola/myr-8-author-vehicle-state-schemamd-canonical-vehiclestate-json](https://github.com/tnando/my-robo-taxi-telemetry/tree/thomasnandola/myr-8-author-vehicle-state-schemamd-canonical-vehiclestate-json) ·

tnando:
**Claude finished @tnando's task in 2m 3s** —— [View job](https://github.com/tnando/my-robo-taxi-telemetry/actions/runs/24226702984)

---
### Review Summary

**Verdict: APPROVE**

**Domain focus:** `agent:sdk-architect`, `agent:go-engineer`

Commit `a4555d5` cleanly addresses the three non-blocking suggestions from the prior review round. All six critical/warning findings from rounds 1 and 2 remain resolved. Good to merge once `contract-guard` is green and MYR-24 is actually filed in Linear.

---

### Verification of new changes (`a4555d5`)

#### 1. Spec-only exemption for §3.1(3) ✅
[`vehicle-state-schema.md` L166](docs/contracts/vehicle-state-schema.md#L166) now reads:
> **Spec-only exemption:** Fields marked `x-spec-only: true` (currently `destinationAddress`, until MYR-24 lands) are exempt from this invariant — they will always be null today regardless of nav state. `contract-tester` MUST skip spec-only fields when evaluating this predicate.

This is exactly the fix I asked for. The `contract-tester` author now has a clear rule to implement, and the predicate's intent is preserved for the post-MYR-24 state. The instruction is also mechanically checkable via `x-spec-only: true` in the schema — no prose parsing needed.

#### 2. `make gen-ts-types` Makefile target ✅
[`Makefile` L41–L48](Makefile#L41-L48) adds:
```makefile
gen-ts-types:
	@command -v npx >/dev/null 2>&1 || { echo "npx not found — install Node.js"; exit 1; }
	npx --yes json-schema-to-typescript@^15 \
		docs/contracts/schemas/vehicle-state.schema.json \
		--output sdk/typescript/src/types/vehicle-state.ts \
		--style.singleQuote --style.semi --additionalProperties false
```
And [§6.1](docs/contracts/vehicle-state-schema.md#L252) now shows `make gen-ts-types` with a note that "CI runs the same target so the doc and the actual command can never drift." Correct framing.

Two non-blocking nits on the target itself (see Suggestions below).

#### 3. `time.Sleep` → `waitForCondition` in `broadcaster_test.go` ✅
[`broadcaster_test.go` L1050–L1056](internal/ws/broadcaster_test.go#L1050-L1056):
```go
// Wait for all 3 route points to be processed (one resolver call each).
// Polling instead of time.Sleep per CLAUDE.md "no sleep in tests" rule.
waitForCondition(t, func() bool {
    resolver.mu.RLock()
    defer resolver.mu.RUnlock()
    return len(resolver.callLog) >= 3
})
```
The `waitForCondition` helper at [L1496](internal/ws/broadcaster_test.go#L1496) already exists with a 2s deadline and 5ms tick — reusing it here is correct, and the `>= 3` condition is the right invariant (one callLog entry per route point). Combined with the existing `callLogBefore` strict-growth wait downstream, the test is now fully deterministic: no sleeps, no off-by-one, no race against `handleDriveEnded`.

This brings the test into compliance with CLAUDE.md's "no sleep in tests" rule. Nicely done.

---

### Suggestions (non-blocking)

1. **`npx` bin name in the Makefile target.** The `json-schema-to-typescript` npm package's CLI bin is named `json2ts`, not `json-schema-to-typescript`. Modern npx (npm ≥ 7) falls back to the single bin when the package name doesn't match a bin, so `npx --yes json-schema-to-typescript@^15 …` will likely resolve to `json2ts` — but it's fragile across npm versions and CI environments. Consider the more explicit form:
   ```makefile
   npx --yes --package=json-schema-to-typescript@^15 json2ts \
       -i docs/contracts/schemas/vehicle-state.schema.json \
       -o sdk/typescript/src/types/vehicle-state.ts \
       --style.singleQuote --style.semi --no-additionalProperties
   ```
   Also note `json2ts` uses `-i`/`-o` (not `--output` as a positional input) and `--no-additionalProperties` (not `--additionalProperties false` as two tokens). Worth validating against the actual `json2ts --help` before CI runs this.

2. **`sdk/typescript/src/types/` path is ahead of the repo.** The output dir doesn't exist yet (no `sdk/` at repo root; only `pkg/sdk/` exists). That's fine since MYR-8 is contract-only, but anyone running `make gen-ts-types` today will hit "directory does not exist." A mkdir guard or a note in the target header would smooth this over:
   ```makefile
   @mkdir -p sdk/typescript/src/types
   ```

3. **Pre-existing `time.Sleep` hygiene pass** (out of scope). `internal/ws/broadcaster_test.go` was the last test in this file carrying a sleep for determinism — now clean. There are still ~6 `time.Sleep` calls elsewhere in that file (timeout-bounded, not determinism-bounded) — worth a sweep in a follow-up to confirm each is necessary. Nothing to do here.

---

### Domain-Specific Notes (sdk-architect)

This PR is in good shape. The contract is honest about its divergences (`speed` outside the GPS atomic group, spec-only fields flagged with `x-spec-only` + `x-spec-only-issue`, §3.1 semantic invariants explicitly not schema-enforceable), and every contract-guard hook the architect needs is mechanically encoded (`x-classification`, `x-encrypted`, `x-atomic-group`, `x-spec-only`). The `contract-tester` author can implement FR/NFR coverage directly from this file without guessing.

---

### Pre-merge checklist (from CLAUDE.md Merge Policy)

- [ ] `contract-guard` passes against `d4ea7de` + `a4555d5`
- [ ] `sdk-architect` final PR review recorded
- [ ] **MYR-24 is actually filed in Linear** and `https://linear.app/myrobotaxi/issue/MYR-24` resolves (this was flagged in the prior round and I couldn't verify it from the PR alone — please confirm before merging, otherwise the "tracked follow-up" is vapor)
- [ ] `golangci-lint run ./...` passes on the branch
- [ ] `go test -race ./...` passes (important for the `broadcaster_test.go` change)



---
 · [Branch: thomasnandola/myr-8-author-vehicle-state-schemamd-canonical-vehiclestate-json](https://github.com/tnando/my-robo-taxi-telemetry/tree/thomasnandola/myr-8-author-vehicle-state-schemamd-canonical-vehiclestate-json)

tnando:
**Claude finished @tnando's task in 1m 39s** —— [View job](https://github.com/tnando/my-robo-taxi-telemetry/actions/runs/24226863731)

---
### Review Summary

**Verdict: APPROVE**

**Domain focus:** `agent:sdk-architect`, `agent:go-engineer`

Commit `d7926bc` is a small, correct follow-up to the previous round's test fix. All six critical/warning findings from rounds 1–2 remain resolved, and this commit actually repairs a latent bug I missed in round 3.

---

### What changed in `d7926bc`

Two small edits:

1. **`internal/ws/route_accumulator.go` L126–L133** — new `Len(vin) int` method that returns `len(a.routes[vin])` under the mutex.
2. **`internal/ws/broadcaster_test.go` L1050–L1057** — replaces the `waitForCondition(len(resolver.callLog) >= 3)` wait with `waitForCondition(b.routes.Len("5YJ3E1EA1NF000001") >= 3)`.

### Why the previous wait was actually broken (missed in round 3)

I want to flag that the round-3 review (including mine) got the invariant wrong. Looking at [`route_broadcast.go` L27–L32](internal/ws/route_broadcast.go#L27-L32):

```go
result := b.routes.Add(payload.VIN, coord)
if !result.ShouldFlush {
    return
}
b.broadcastRoutePoints(ctx, event.ID, payload.VIN, result.Points)
```

`handleDriveUpdated` only calls `broadcastRoutePoints` (and therefore the resolver) **when the batch flushes**. The test configures `newRouteAccumulator(100, 0)` at L1022 — batch size 100, zero flush interval — specifically to prevent flushing. So with only 3 route points, `ShouldFlush` is always false, the resolver is **never called**, and `resolver.callLog` stays at 0. The prior `waitForCondition(len(resolver.callLog) >= 3)` would have hit its 2 s deadline and failed the test.

The `d7926bc` fix is therefore not just a cleanup — it's repairing a broken wait. The new `Len()` method directly measures the invariant the test actually cares about (route points landed in the accumulator) without coupling to the flush-triggered resolver side effect. Correct call.

### Code quality

- `Len()` uses the same `sync.Mutex` Lock pattern as the rest of the file (consistent with `Add`, `Flush`, `Clear`). ✅
- Doc comment is present, explains intent and the "no state mutation" contract. ✅
- Private method on a private type, only reachable from same-package tests — no API surface leak. ✅
- No new allocations, no goroutines, no channels. ✅
- Comment block on the test wait now explains *why* we poll the accumulator rather than resolver calls, which was the missing context a future maintainer would need. ✅

### Race detector

`Len()` holds the mutex for the entire `len()` read, and no other method holds the mutex across long operations. `go test -race ./internal/ws/...` should pass on this change. Worth confirming in CI, but I don't see a data-race hazard by inspection.

### Pre-existing state (unchanged by this commit, but still relevant)

Per the round-4 merge checklist, these still need to hold before merge:

- [ ] `contract-guard` green against `d4ea7de` + `a4555d5` + `d7926bc`
- [ ] `sdk-architect` final PR review recorded
- [ ] **MYR-24 actually filed in Linear** — `https://linear.app/myrobotaxi/issue/MYR-24` must resolve, otherwise the "spec-only, tracked follow-up" annotations in `vehicle-state-schema.md` §1.1 and §7.2 are dangling references
- [ ] `golangci-lint run ./...` passes
- [ ] `go test -race ./...` passes (particularly `TestBroadcaster_DriveEndedClearsAccumulator`)

### Suggestions (non-blocking)

- **Makefile target `gen-ts-types` still uses the wrong CLI invocation** (raised in round 4, not addressed here). `json-schema-to-typescript` exposes a bin named `json2ts` with `-i`/`-o` flags, not `--output` as a positional. Worth validating against `npx --yes --package=json-schema-to-typescript@^15 json2ts --help` before CI actually runs it. Out of scope for this commit but still open.
- **Output directory `sdk/typescript/src/types/` doesn't exist yet.** A `@mkdir -p` guard in the Makefile target would make it idempotent for anyone running it today. Out of scope, but easy follow-up.

---

### Domain-Specific Notes (go-engineer + sdk-architect)

**go-engineer:** This is idiomatic Go. A 7-line accessor method with a mutex, a doc comment, and a test-only motivation is exactly the right shape — no over-engineering, no exposing internals beyond what the test genuinely needs. The `Len()` approach is also strictly more correct than reaching into `b.routes.routes` via reflection or exposing the map, which would have been the alternatives.

**sdk-architect:** No contract surface changes in this commit — `Len()` is private (lowercase), on a private type (`routeAccumulator`), and the public broadcast shape is untouched. Contract-guard should be indifferent to this commit.

---

This PR is good to merge once `contract-guard`, `golangci-lint`, `go test -race`, and the MYR-24 Linear-issue existence check all pass.



--- · [Branch: thomasnandola/myr-8-author-vehicle-state-schemamd-canonical-vehiclestate-json](https://github.com/tnando/my-robo-taxi-telemetry/tree/thomasnandola/myr-8-author-vehicle-state-schemamd-canonical-vehiclestate-json)

tnando:
**Claude finished @tnando's task in 1m 19s** —— [View job](https://github.com/tnando/my-robo-taxi-telemetry/actions/runs/24226985914)

---
### Review Summary

**Verdict: APPROVE**

**Domain focus:** `agent:sdk-architect`, `agent:go-engineer`

Commit `0fdab2e` cleanly lands both non-blocking suggestions from the round-4 review. All six critical/warning findings from rounds 1–2 remain resolved, and the `broadcaster_test.go` / `route_accumulator.go` changes from rounds 3–5 remain correct. This PR is good to merge once CI is green.

---

### Verification of `0fdab2e`

The Makefile `gen-ts-types` target now matches the `json2ts` CLI contract exactly:

```makefile
gen-ts-types:
    @command -v npx >/dev/null 2>&1 || { echo "npx not found — install Node.js"; exit 1; }
    @mkdir -p sdk/typescript/src/types
    npx --yes --package=json-schema-to-typescript@^15 json2ts \
        -i docs/contracts/schemas/vehicle-state.schema.json \
        -o sdk/typescript/src/types/vehicle-state.ts \
        --no-additionalProperties
    @echo "Generated sdk/typescript/src/types/vehicle-state.ts"
```

What changed:

1. **`--package=…@^15 json2ts` explicit form** ✅ — no longer relies on the npx-7+ fallback of resolving a mismatched package-name/bin-name. Robust across npm versions and CI environments.
2. **`-i` / `-o` flags** ✅ — these are the canonical `json2ts` CLI arguments. Previous version used `--output` + a positional input which is not how the `json-schema-to-typescript` CLI actually parses its argv.
3. **`--no-additionalProperties`** ✅ — correct single-token flag form. Previous `--additionalProperties false` would have been parsed as two separate tokens.
4. **`@mkdir -p sdk/typescript/src/types`** ✅ — idempotent guard so the target works on a fresh checkout before the `sdk/typescript/` tree exists (which it doesn't yet in this repo — `pkg/sdk/` is the only `sdk` path today).

This resolves both non-blocking suggestions from round 4.

### Nit (non-blocking, for a follow-up)

The round-4 version passed `--style.singleQuote --style.semi` and this commit drops them, so generated output will use double quotes and default formatting. If the `@myrobotaxi/sdk` TS package ends up configured with single-quote style, these flags will need to come back (or the generated file will conflict with Prettier/eslint on first commit). Easy fix when MYR-24 or the TS SDK work lands; nothing to do in this PR.

---

### Status of all prior-round findings

| # | Finding | Status |
|---|---------|--------|
| 1 | `required` array missing non-nullable fields | ✅ Resolved in `d4ea7de` (option b — spec-only + MYR-24) |
| 2 | Schema vs `internal/store/types.go` gap | ✅ Documented in §1.1, §7.2, tracked under MYR-24 |
| 3 | §3.1 non-null predicates not schema-enforceable | ✅ Enforcement-boundary paragraph added; predicates tagged "semantic invariant only" |
| 4 | Unreachable `$defs` sub-schemas | ✅ Deleted; §5.2 documents the deliberate absence |
| 5 | `latitude`/`longitude` description contradicts type | ✅ Rewritten to match "0,0 = no fix" convention |
| 6 | `chargeLevel: 0` semantics disagree | ✅ Schema annotation now matches §2.2 verbatim |
| Rd3 | `broadcaster_test.go` flake fix | ✅ `callLogBefore` strict-growth wait |
| Rd4 | `time.Sleep` in broadcaster test | ✅ Replaced with `waitForCondition` |
| Rd4 | §3.1 spec-only exemption | ✅ Added |
| Rd4 | Makefile `gen-ts-types` target | ✅ Added (round 4) + fixed (round 5) |
| Rd5 | `routeAccumulator.Len()` for accurate test wait | ✅ Fixed the broken flush-dependent `callLog` invariant |

---

### Domain-Specific Notes

**sdk-architect:** Contract is mechanically complete and honest. Every drift hook `contract-guard` needs is encoded (`x-classification`, `x-encrypted`, `x-atomic-group`, `x-spec-only`, `x-spec-only-issue`). Divergences (`speed` outside GPS group, spec-only fields) are documented with rationale. `contract-tester` can implement FR/NFR coverage directly from this file without guessing.

**go-engineer:** The test-path changes (`broadcaster_test.go` + `route_accumulator.Len()`) are idiomatic Go — 7-line mutex-guarded accessor, polling helper reuse, no sleeps, no globals, no goroutine leaks. Nothing to flag.

---

### Pre-merge checklist (from CLAUDE.md Merge Policy)

- [ ] `contract-guard` green against `d4ea7de` + `a4555d5` + `d7926bc` + `0fdab2e`
- [ ] `sdk-architect` final PR review recorded
- [ ] **MYR-24 actually filed in Linear** — `https://linear.app/myrobotaxi/issue/MYR-24` must resolve; the "spec-only, tracked follow-up" annotations in §1.1 and §7.2 become dangling references otherwise
- [ ] `golangci-lint run ./...` passes
- [ ] `go test -race ./...` passes (particularly `TestBroadcaster_DriveEndedClearsAccumulator`)
- [ ] `go build ./cmd/...` passes

With those green, this is good to merge.



---
 · [Branch: thomasnandola/myr-8-author-vehicle-state-schemamd-canonical-vehiclestate-json](https://github.com/tnando/my-robo-taxi-telemetry/tree/thomasnandola/myr-8-author-vehicle-state-schemamd-canonical-vehiclestate-json) ·

PR #178, opened 2026-04-20
MYR-25 Add charge field instrumentation for Tesla unit verification

tnando:
## Summary

- Adds `TimeToFullCharge` (proto 43) and `EstimatedHoursToChargeTermination` (proto 190) to field map + fleet config at 30s intervals
- Adds temporary debug logging in the decoder for the 3 fields under verification (MYR-25/28/29)
- No wire mapping, DB columns, or broadcaster changes — observation only

## Why

The v1 contract assumes `TimeToFullCharge` is in **hours** (decimal) based on secondary sources, but no empirical protobuf capture has confirmed this. This PR instruments the decoder so we can compare raw values against the Tesla touchscreen display while charging and navigating.

## What to do after merge

1. Deploy to Fly.io
2. Push updated fleet config: `./scripts/push-fleet-config.sh --vin <VIN> --token <TOKEN>`
3. Plug in Tesla to charge + set a nav destination
4. Read `fly logs` — look for `MYR-25/28/29 FIELD VERIFICATION` lines
5. Compare raw values against Tesla touchscreen → ground truth for units

## Cleanup

The debug logging is temporary. After verification, a follow-up commit removes the `switch` block in `decoder.go` and records the findings in the contract docs.

## Test plan

- [x] `go test ./...` — all pass
- [x] `go vet ./...` — clean
- [x] `go build ./cmd/...` — builds
- [x] `TestDefaultFieldConfig_CoversAllTrackedFields` — passes (new fields have config entries)
- [ ] CI checks pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #179, opened 2026-04-20
MYR-25 Fix gocognit lint failure in DecodePayload

tnando:
## Summary

Extracts the MYR-25/28/29 verification debug logging from `DecodePayload` into two helper functions (`logVerificationField`, `logFieldVerification`) to bring cognitive complexity from 22 back under the 20 threshold.

## Root cause

The debug `switch` + `if` blocks added in PR #178 pushed `DecodePayload` over the `gocognit` limit. CI lint caught it on the main branch merge.

## Test plan

- [x] `golangci-lint run ./internal/telemetry/` — 0 issues
- [x] `go test ./...` — all pass
- [ ] CI checks pass

### DIFF
--- a/internal/telemetry/decoder.go
+++ b/internal/telemetry/decoder.go
@@ -107,25 +107,8 @@ func (d *Decoder) DecodePayload(payload *tpb.Payload) (events.VehicleTelemetryEv
 			continue
 		}
 
-		// MYR-25/28/29: Temporary debug log for Tesla field unit verification.
-		// These fields are NOT in fieldMap (intentionally excluded from the event
-		// bus to avoid leaking uncontracted fields to WS clients). We intercept
-		// them here for observation only. Remove after empirical verification.
-		switch datum.GetKey() {
-		case tpb.Field_TimeToFullCharge, tpb.Field_EstimatedHoursToChargeTermination:
-			if tv, err := extractValue(datum); err == nil {
-				vin := payload.GetVin()
-				vinSuffix := vin
-				if len(vin) > 4 {
-					vinSuffix = vin[len(vin)-4:]
-				}
-				slog.Info("MYR-25/28/29 FIELD VERIFICATION",
-					slog.String("field", datum.GetKey().String()),
-					slog.String("vin_last4", vinSuffix),
-					slog.Any("raw_value", tv),
-				)
-			}
-		}
+		// MYR-25/28/29: observation-only log for untracked charge fields.
+		logVerificationField(datum, payload.GetVin())
 
 		name, ok := InternalFieldName(datum.GetKey())
 		if !ok {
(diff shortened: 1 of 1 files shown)


tnando:
**Claude finished @tnando's task in 53s** —— [View job](https://github.com/tnando/my-robo-taxi-telemetry/actions/runs/24647935870)

---
## Review Summary

**Verdict: APPROVE**

**Domain focus:** none (general Go review only)

### Findings

#### Critical (must fix)
- None.

#### Warnings (should fix)
- None.

#### Suggestions
- **Minor: redundant `extractValue` call on the `MilesToArrival` path.** `logVerificationField` is invoked on every datum before the `InternalFieldName` check, and for `Field_TimeToFullCharge` / `Field_EstimatedHoursToChargeTermination` it calls `extractValue(datum)`. These two fields are not in `fieldMap`, so the `continue` at line 115 fires for them — no double extraction in practice. Behavior is preserved exactly. No change needed, but worth being aware if either key ever gets added to `fieldMap`: you'd then call `extractValue` twice per datum. A small guard (e.g. only call the helper when the field is untracked) would be a cheap future-proofing, but it's not a blocker for a temporary debug log.
- The two helpers are file-private and well-scoped — keeping them in `decoder.go` is the right call given the `// Remove after empirical verification is complete.` comment. Suggest adding a brief tracking task / TODO with a target removal date so the debug code doesn't outlive its purpose.

### Domain-Specific Notes

Behavior-preserving refactor of existing debug logging:
- ✅ VIN redaction preserved (last-4 pattern in `logFieldVerification`) — no regression against the "VINs not logged in full" rule.
- ✅ `slog.Info` with structured fields — matches project logging conventions.
- ✅ No error swallowed: the `if err == nil` guard in `logVerificationField` mirrors the original inline code; since this is observation-only logging on untracked fields, silently skipping on extract error is correct (the real decode path emits `FieldError` elsewhere).
- ✅ Helper doc comments are clear, name the tracking tickets, and flag the code as temporary.
- ✅ File size: 234 lines — well under the 300-line limit.
- ✅ No new deps, no panics, no globals, no exported API changes — pure internal code motion to reduce `gocognit`.
- ✅ Call sites read more clearly than the previous inline switch.
- ✅ Commit message follows `MYR-<num> <Imperative verb> ...` convention. Small, focused diff (+30/-31, single file). Author reports local `golangci-lint run ./internal/telemetry/` and `go test ./...` clean.

Tests weren't added or changed — appropriate here since the refactor is behavior-preserving and the existing decoder tests cover `DecodePayload`. If `TestDecoder_*` already passes against this branch, no additional coverage is required for a pure extract-method change.

LGTM.


 • Branch: [`fix/myr-25-lint-complexity`](https://github.com/tnando/my-robo-taxi-telemetry/tree/fix/myr-25-lint-complexity)

PR #183, opened 2026-04-21
MYR-37 Add ops auth link (Tesla OAuth authorization_code + PKCE)

tnando:
## Summary
Adds `ops auth link --user-id <id>` — a full Tesla OAuth2 authorization_code + PKCE flow directly in the CLI. Closes the recovery gap when the stored `refresh_token` is revoked/expired and `ops auth token` fails with `401 login_required`.

## Flow

1. Generate PKCE verifier/challenge (S256) per run
2. Spin up ephemeral `127.0.0.1:<port>/callback` HTTP server
3. Open Tesla's `/oauth2/v3/authorize` URL in the default browser
4. Validate returned `state`; surface Tesla-reported errors
5. Exchange code for tokens via `/oauth2/v3/token`
6. Persist fresh `access_token`/`refresh_token`/`expires_at` via `AccountRepo.UpdateTeslaToken`

## One-time developer setup
Register `http://localhost:8765/callback` (or `--port` override) on the Tesla Fleet API app's allowed redirect URIs. Documented in `docs/ops-cli.md`.

## Test plan
- [x] `go vet ./...`
- [x] `go test ./...`
- [x] Unit tests cover PKCE generation (verifier/challenge SHA-256 relation), authorize URL param shape, callback handler (success + state mismatch + Tesla-reported error), and token exchange form body
- [ ] Manual: `ops auth link --user-id <id>` against a real Tesla account, confirm `ops auth token` succeeds after
- [ ] Manual: confirm port collision error message is clear if port not registered on Tesla app

Closes MYR-37.

### DIFF
--- a/cmd/ops/auth.go
+++ b/cmd/ops/auth.go
@@ -20,11 +20,13 @@ var errRefreshSkipped = errors.New("tesla token refresh skipped")
 
 func runAuth(ctx context.Context, args []string) error {
 	if len(args) == 0 {
-		return fmt.Errorf("auth requires a subcommand (token)")
+		return fmt.Errorf("auth requires a subcommand (token | link)")
 	}
 	switch args[0] {
 	case "token":
 		return runAuthToken(ctx, args[1:])
+	case "link":
+		return runAuthLink(ctx, args[1:])
 	default:
 		return fmt.Errorf("unknown auth subcommand %q", args[0])
 	}
--- a/cmd/ops/auth_link.go
+++ b/cmd/ops/auth_link.go
@@ -2,23 +2,16 @@ package main
 
 import (
 	"context"
-	"crypto/rand"
-	"crypto/sha256"
-	"encoding/base64"
-	"encoding/json"
 	"errors"
 	"flag"
 	"fmt"
 	"html"
-	"io"
 	"log/slog"
 	"net"
 	"net/http"
-	"net/url"
 	"os"
 	"os/exec"
 	"runtime"
-	"strings"
 	"time"
 
 	"github.com/tnando/my-robo-taxi-telemetry/internal/store"
@@ -118,53 +111,9 @@ func runAuthLink(ctx context.Context, args []string) error {
 
 	return writeJSON(os.Stdout, authLinkOutput{
 		UserID:    *userID,
-		ExpiresAt: expiresAt.UTC().Format(time.RFC3339),
+		ExpiresAt: formatExpiry(expiresAt),
 		Message:   "Tesla account linked — run `ops auth token` to verify",
 	})
-}
-
-// pkcePair holds a PKCE verifier/challenge pair for a single OAuth flow.
-type pkcePair struct {
-	verifier  string
-	challenge string
-}
-
-// newPKCE generates a fresh PKCE verifier + S256 challenge per RFC 7636.
-// The verifier is a URL-safe random string; the challenge is
-// base64url(sha256(verifier)) without padding.
-func newPKCE() (pkcePair, error) {
-	verifier, err := randomURLSafeString(32)
-	if err != nil {
-		return pkcePair{}, err
-	}
-	sum := sha256.Sum256([]byte(verifier))
-	challenge := base64.RawURLEncoding.EncodeToString(sum[:])
-	return pkcePair{verifier: verifier, challenge: challenge}, nil
-}
-
-// randomURLSafeString returns n bytes of cryptographic randomness encoded
-// with unpadded base64url, producing a string safe to use in URLs and
-// query parameters.
-func randomURLSafeString(n int) (string, error) {
-	b := make([]byte, n)
-	if _, err := rand.Read(b); err != nil {
-		return "", fmt.Errorf("rand.Read: %w", err)
-	}
-	return base64.RawURLEncoding.EncodeToString(b), nil
-}
-
-// buildAuthorizeURL constructs the Tesla /oauth2/v3/authorize URL that
-// starts the authorization_code + PKCE flow.
-func buildAuthorizeURL(clientID, redirectURI, scopes, state, codeChallenge string) string {
-	q := url.Values{}
-	q.Set("response_type", "code")
-	q.Set("client_id", clientID)
-	q.Set("redirect_uri", redirectURI)
-	q.Set("scope", scopes)
-	q.Set("state", state)
-	q.Set("code_challenge", codeChallenge)
-	q.Set("code_challenge_method", "S256")
-	return teslaOAuthAuthorizeURL + "?" + q.Encode()
 }
 
 // runCallbackServer starts a one-shot HTTP server on 127.0.0.1:<port>,
@@ -234,6 +183,8 @@ type callbackResult struct {
 // callbackHandler returns an http.HandlerFunc that validates the state
 // parameter, surfaces Tesla-reported errors, and forwards the auth code
 // through the result channel. The page the user sees reflects the outcome.
+// Non-blocking sends protect against duplicate callbacks (double-click,
+// refresh) holding handler goroutines open forever.
 func callbackHandler(expectedState string, result chan<- callbackResult) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		q := r.URL.Query()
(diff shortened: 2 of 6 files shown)


tnando:
**Claude finished @tnando's task in 1m 41s** —— [View job](https://github.com/tnando/my-robo-taxi-telemetry/actions/runs/24751777434)

---
## Review Summary

**Verdict: COMMENT**

**Domain focus:** none (general review applied)

Nice work — the PKCE flow, state validation, HTML escaping on the callback page, 127.0.0.1-only listener, body-size limit on the token response, and clean error-path shutdown are all on-point. A few issues worth addressing before merge, none of them blockers on their own.

### Findings

#### Critical (must fix)
_None._

#### Warnings (should fix)

- **`cmd/ops/auth_link.go` is 357 lines — exceeds the 300-line cap in CLAUDE.md** ("Max 300 lines per file (excluding tests). Decompose if exceeded"). Natural split points: move the OAuth primitives (`newPKCE`, `randomURLSafeString`, `buildAuthorizeURL`, `exchangeCodeForToken`, `tokenResponse`) into `auth_oauth.go`, or lift the callback-server machinery (`runCallbackServer`, `callbackHandler`, `respondCallback`, `openBrowser`) into `auth_link_callback.go`. Either leaves `runAuthLink` as the thin orchestrator in `auth_link.go`.

- **`exchangeCodeForToken` uses `http.DefaultClient` (no timeout)** — `cmd/ops/auth_link.go:331`. The sibling `internal/telemetry/token_refresher.go:52` already uses `&http.Client{Timeout: 15 * time.Second}` for the same endpoint. Parent-context cancellation covers user `Ctrl-C`, but a hung TCP connection to `auth.tesla.com` will hold up the CLI until the OS TCP timeout (minutes). Use a local client with an explicit timeout to match the existing pattern.

- **The `TestExchangeCodeForToken_*` test doesn't actually call `exchangeCodeForToken`** — `cmd/ops/auth_link_test.go:136-197`. It stands up a test server, then builds its own `http.NewRequestWithContext`/`srv.Client().Do` and asserts on that. The comment at line 150-153 even acknowledges this ("we can't mock it without touching the constant"). Net effect: the production function's form assembly, response parsing, status handling, empty-token guard, and error-wrapping are untested. Two straightforward fixes, in order of preference:
  1. Make the endpoint injectable — either a package-level `var teslaOAuthTokenURL = "..."` (swap in the test) or a `tokenEndpoint string` parameter on `exchangeCodeForToken`. Then call the real function against `srv.URL` and assert on its `*tokenResponse` return value.
  2. Extract the form assembly into `buildTokenRequestForm(...) url.Values` and test that directly.

  Option 1 is better because it also exercises the non-200 branch and the "missing access_token/refresh_token" guard at line 353.

#### Suggestions

- **`result` channel could block a second callback handler forever.** `cmd/ops/auth_link.go:187` allocates `make(chan callbackResult, 1)`. If the browser fires `/callback` twice (user double-clicks, or a retry), the second handler's `result <- ...` blocks indefinitely. `server.Shutdown` waits 2s for in-flight handlers before hard-cancelling, so in practice the process exits cleanly — but a non-blocking `select { case result <- ...: default: }` in `callbackHandler` would be more defensive.

- **`openBrowser` uses `exec.CommandContext(ctx, ...)`** — when `runAuthLink`'s context is cancelled, the browser subprocess gets killed. Probably fine (we're shutting down anyway), but slightly surprising for `open`/`xdg-open` which normally detach. `exec.Command` (unbound) would be equally correct here since the process self-detaches.

- **Logging the raw token-endpoint error body** at `cmd/ops/auth_link.go:344` is fine for a dev CLI, but Tesla error responses occasionally echo request parameters. Worth a `slog.String("body", truncate(body, 512))` if you ever surface this beyond developer machines.

- **Naming nit:** `formatExpiry` (in `auth.go`) and the ad-hoc `expiresAt.UTC().Format(time.RFC3339)` at `auth_link.go:121` duplicate each other. The latter could reuse `formatExpiry(expiresAt)` for consistency.

### Domain-Specific Notes
No agent labels on this PR — applied general Go + CLAUDE.md rules only.

---

**Verdict: COMMENT.** The file-size cap and the non-exercising test are the two items I'd expect fixed before merge; the `http.DefaultClient` timeout is a close third. The flow itself is correctly implemented.
 · [branch](https://github.com/tnando/my-robo-taxi-telemetry/tree/thomasnandola/myr-37-add-ops-auth-link-for-tesla-oauth-authorization_code-flow)

tnando:
**Claude finished @tnando's task in 1m 43s** —— [View job](https://github.com/tnando/my-robo-taxi-telemetry/actions/runs/24752047348)

---
### Review Summary

**Verdict: APPROVE**

**Domain focus:** none (general Go + CLAUDE.md review applied)

The second commit (`3a2fbfb`) addresses every warning from the prior review. The OAuth flow implementation is clean, testable, and security-aware.

### Prior feedback — all resolved

| Prior warning | Status | Evidence |
|---|---|---|
| `auth_link.go` was 357 lines (>300 cap) | ✅ Fixed | Split into `auth_link.go` (266) + `auth_oauth.go` (137) — orchestration vs. OAuth primitives |
| `exchangeCodeForToken` used `http.DefaultClient` (no timeout) | ✅ Fixed | `auth_oauth.go:22,96` — `tokenExchangeTimeout = 15s` in a per-call `&http.Client{}` |
| `TestExchangeCodeForToken_*` didn't exercise the real function | ✅ Fixed | Package-level `var teslaOAuthTokenEndpoint` (`auth_oauth.go:27`) swapped via `withTokenEndpoint` (`auth_link_test.go:254`). Real function is now exercised through `httptest.Server` for 200-parse, 401-error, and empty-token branches |
| Duplicate `/callback` could block on the `result` channel | ✅ Fixed | `trySendResult` non-blocking send (`auth_link.go:220-225`) + regression test `TestCallbackHandler_DuplicateSendDoesNotBlock` |
| `formatExpiry` duplication on output | ✅ Fixed | `auth_link.go:114` now calls `formatExpiry(expiresAt)` |
| gosec G204 subprocess findings | ✅ Fixed | Per-call `//#nosec G204` with clear rationale; URL is CLI-built from static constants, not user input |

### Findings

#### Critical
_None._

#### Warnings
_None._

- **`openBrowser` still uses `exec.CommandContext(ctx, ...)`** (`auth_link.go:252-256`). On Ctrl-C the subprocess gets killed — harmless since we're exiting, but `exec.Command` (unbound) matches the self-detaching semantics most `open`/`xdg-open` users expect. Left as-is is fine.
- **Callback HTTP server has only `ReadHeaderTimeout` set.** For a 2-minute-lifetime localhost-only listener this is acceptable, but a `WriteTimeout` would be defensive (`auth_link.go:143`).
- **Token-error body logging** (`auth_oauth.go:107-112`) is bounded implicitly by the 64KB `io.LimitReader`. If this ever surfaces beyond developer machines, consider an explicit `truncate(body, 512)` since Tesla error responses occasionally echo parameters.

### Quality check (against CLAUDE.md)

- ✅ File-size cap: all changed files ≤ 266 lines
- ✅ Errors wrapped with `fmt.Errorf("context: %w", err)` throughout
- ✅ Context propagation on every I/O call
- ✅ No globals in prod paths (the one `var teslaOAuthTokenEndpoint` is a documented, scoped test seam)
- ✅ No `panic()`, no `any`/`interface{}`
- ✅ `nhooyr.io/websocket` compliance N/A (no WebSocket changes)
- ✅ Secrets from env vars (`AUTH_TESLA_ID`/`AUTH_TESLA_SECRET`)
- ✅ Client secret never reaches the authorize URL (PKCE-only) — only the token endpoint
- ✅ `t.Helper()` + `t.Cleanup()` used correctly in `withTokenEndpoint`
- ✅ Listener bound to `127.0.0.1:<port>` (not `0.0.0.0`) — `auth_link.go:131`
- ✅ HTML escape on callback page output — `auth_link.go:232-233`
- ✅ Response body bounded by `io.LimitReader(resp.Body, 1<<16)` — `auth_oauth.go:103`
- ✅ State parameter validated (CSRF protection) — `auth_link.go:199-203`
- ✅ PKCE S256 per RFC 7636 — `auth_oauth.go:38-46`

### Domain-Specific Notes
No `Agent/*` labels on this PR, so general Go + CLAUDE.md rules only. The OAuth flow is self-contained CLI scaffolding — no WebSocket/DB schema/contract touchpoints, so no `sdk-architect` or `contract-guard` involvement required.

---

Nice iteration. The test seam via `teslaOAuthTokenEndpoint` is the right trade-off here — one narrowly-scoped package-level var in exchange for exercising the real function's parse/status/empty-field branches. Ready to merge.


 · [branch](https://github.com/tnando/my-robo-taxi-telemetry/tree/thomasnandola/myr-37-add-ops-auth-link-for-tesla-oauth-authorization_code-flow)

PR #184, opened 2026-04-22
MYR-38 Let DEBUG_FIELDS_TOKEN gate /api/debug/fields in prod

tnando:
## Summary
Decouples `/api/debug/fields` and `PublishRawFields` from the `--dev` flag. In MYR-36 I tied three concerns to one flag — only loose-auth belongs under `--dev`. This change lets operators enable the debug endpoint on the production server via `DEBUG_FIELDS_TOKEN`, so `ops fields watch` can finally stream **real-Tesla** frames (the motivating use case of MYR-36, which wasn't deliverable under the original scoping).

## Gate matrix

| `--dev` | `DEBUG_FIELDS_TOKEN` | Endpoint mounted? | Client token required? |
|---|---|---|---|
| false | empty | no | — |
| false | < 32 chars | **startup fails** | — |
| false | ≥ 32 chars | yes | yes |
| true | empty | yes | no |
| true | any | yes | yes (enforced) |

Main-WS JWT auth semantics are **unchanged** — `--dev` still bypasses it and never otherwise.

## Enablement on prod
```bash
flyctl secrets set DEBUG_FIELDS_TOKEN=\"\$(openssl rand -base64 32)\" -a myrobotaxi-telemetry
```
Save the value; you cannot read it back from Fly.

## Test plan
- [x] `go vet ./...`
- [x] `go test ./...`
- [x] New `TestResolveDebugFieldsGate` covers all five matrix rows, including the ≥32 char enforcement error path
- [x] `docs/ops-cli.md` rewritten: gate matrix, prod enablement, local-vs-prod streaming, new troubleshooting entries
- [ ] Manual on prod: set the secret, run `ops fields watch --server wss://telemetry.myrobotaxi.app --vin <vin>`, confirm frames appear

Closes MYR-38. Unblocks all future field-verification issues (MYR-25/28/29-style) — they now have a one-command validation tool that hits real data.

### DIFF
--- a/cmd/telemetry-server/adapters.go
+++ b/cmd/telemetry-server/adapters.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"crypto/tls"
+	"errors"
 	"fmt"
 	"log/slog"
 	"net"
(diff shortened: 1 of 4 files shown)


tnando:
**Claude finished @tnando's task in 1m 56s** —— [View job](https://github.com/tnando/my-robo-taxi-telemetry/actions/runs/24752808601)

---
### Review Summary

**Verdict: APPROVE**

**Domain focus:** none (applied general Go review + CLAUDE.md rules)

Clean refactor that unbundles the conflated concerns of MYR-36: `--dev` now gates only loose auth for the main WS, while `DEBUG_FIELDS_TOKEN` independently gates `/api/debug/fields` + `PublishRawFields`. The gate matrix in the PR description matches the code exactly, and the test covers every row.

### Findings

#### Critical (must fix)
None.

#### Warnings (should fix)
None.

#### Suggestions
- **`PublishRawFields` now on in prod whenever the token is set.** The old comment called it out as "extra work" kept off in prod on purpose. The new path trades that for on-demand debugging without a redeploy, which is the right call — but worth noting that a prod instance with `DEBUG_FIELDS_TOKEN` set pays the raw-publish cost even when no debug client is connected. For the current telemetry volume (~10 msgs/sec/vehicle) this is negligible; if the fleet grows, consider lazy-enabling the raw publisher only while at least one debug-fields subscriber is attached. Not a blocker — just a future optimization hook.
- **Token comparison in `DebugFieldsHandler.authorize` uses `v == h.cfg.APIKey`** (non-constant-time). Pre-existing, not introduced by this PR, but now that this endpoint is reachable on the public production hostname it's a better target than it was under `--dev`-only. Consider `subtle.ConstantTimeCompare` in a follow-up. ([Fix this →](https://claude.ai/code?q=In%20internal%2Ftelemetry%2Fdebug_fields_handler.go%20authorize%28%29%2C%20replace%20the%20%60v%20%3D%3D%20h.cfg.APIKey%60%20string%20equality%20checks%20on%20the%20X-Debug-Token%20header%20and%20%3Ftoken%3D%20query%20param%20with%20subtle.ConstantTimeCompare%20to%20mitigate%20timing%20side%20channels%20on%20the%20production-exposed%20endpoint.%20Branch%3A%20thomasnandola%2Fmyr-38-decouple-apidebugfields-from-dev-so-it-can-be-enabled-in&repo=tnando/my-robo-taxi-telemetry))
- **Minor doc/code drift risk:** `debugFieldsMinTokenLen` is a constant, but the error string `"DEBUG_FIELDS_TOKEN must be at least 32 chars in non-dev mode"` hardcodes `32`. If the const ever changes, the message becomes a lie. Consider `fmt.Sprintf` or composing the error at construction time. Very minor.
- **`cmd/telemetry-server/main.go` is 381 lines** (CLAUDE.md says ≤300). Pre-existing with `//nolint:funlen` on `run()`, and this PR only adds ~20 net lines to the composition root. Not a blocker, but the composition root is getting long enough that extracting `setupDebugFieldsEndpoint(...)` (mirroring `setupFleetConfigEndpoint`) would tidy it up.

### General checks
- ✅ Error handling wraps with `%w`; `errors.Is` used correctly (incl. `resolveDebugFieldsGate` returning a wrapped sentinel that the test matches via `errors.Is`)
- ✅ Named exports have doc comments; unexported helpers are also documented
- ✅ No panics, no new globals, no `init()`
- ✅ Dependency direction preserved (`cmd/ → internal/telemetry`)
- ✅ Secret via env var, never in code/config
- ✅ Table-driven test with `t.Run`, covers all 5 matrix rows; `strings.Repeat("a", debugFieldsMinTokenLen)` exercises the exact-boundary case
- ✅ `go vet`, `go test`, `golangci-lint` called out as green in the PR test plan
- ✅ Startup logs the active gate reason and whether a token is required — good operator signal
- ✅ Entropy math in the comment (32 base64 chars ≈ 192 bits) is accurate, matches the `openssl rand -base64 32` example in the docs (which actually yields 44 chars / 256 bits — comfortably above the floor)

### Domain-Specific Notes
No agent labels on this PR, so only general review applied. That said, a light security lens is warranted since this flips an endpoint from dev-only to reachable-in-prod:

- mTLS vehicle ingress path is untouched.
- Main-WS JWT auth semantics are explicitly preserved (`--dev` still the only bypass, and this PR doesn't touch that branch).
- The new endpoint requires a shared secret on prod and rejects weak secrets at startup — good fail-fast behavior.
- Docs correctly warn operators they won't be able to read the token back from Fly and must save it locally.

### Todo
- [x] Check CLAUDE.md rules against changes
- [x] Apply general Go review
- [x] Post verdict
 · [Branch](https://github.com/tnando/my-robo-taxi-telemetry/tree/thomasnandola/myr-38-decouple-apidebugfields-from-dev-so-it-can-be-enabled-in)

PR #185, opened 2026-04-22
MYR-39 Fix openssl cert recipe + commit dev-notls config

tnando:
## Summary
Two small onboarding unblocks bundled into one docs+config PR:

1. **`docs/simulator.md` cert recipe** was broken on Homebrew openssl@3 ≥3.6 (it looks for `/opt/homebrew/etc/openssl@3/openssl.cnf` which is no longer shipped). Fix: wrap the block in `OPENSSL_CONF=/dev/null` so openssl falls back to compiled-in defaults. Verified end-to-end against the real `certs/` directory; server cert carries `subjectAltName: DNS:localhost, IP:127.0.0.1` as expected.
2. **`configs/dev-notls.json`** is referenced by `docs/ops-cli.md` (from #184) but was never actually committed. This PR adds the file so the "local dev server without TLS certs" onboarding path actually works.

## Test plan
- [x] Ran the updated cert block verbatim from a clean shell on macOS + Homebrew openssl@3 3.6.1 — all commands succeed
- [x] Confirmed server cert SAN contains `DNS:localhost, IP:127.0.0.1`
- [x] Confirmed client cert CN is the expected VIN
- [x] `configs/dev-notls.json` matches the structure the composition root in `cmd/telemetry-server/main.go` expects

Docs/config only; no Go touched.

Closes MYR-39.

### DIFF
--- a/configs/dev-notls.json
+++ b/configs/dev-notls.json
@@ -1,0 +1,38 @@
+{
+  "server": {
+    "tesla_port": 8443,
+    "client_port": 8080,
+    "metrics_port": 9090
+  },
+  "tls": {
+    "cert_file": "",
+    "key_file": "",
+    "ca_file": ""
+  },
+  "database": {
+    "max_conns": 10,
+    "min_conns": 2
+  },
+  "telemetry": {
+    "max_vehicles": 100,
+    "event_buffer_size": 1000,
+    "batch_write_interval": "5s",
+    "batch_write_size": 100
+  },
+  "drives": {
+    "min_duration": "2m",
+    "min_distance_miles": 0.1,
+    "end_debounce": "30s",
+    "geocode_timeout": "5s"
+  },
+  "websocket": {
+    "heartbeat_interval": "15s",
+    "write_timeout": "10s",
+    "max_connections_per_user": 5,
+    "read_limit": 4096
+  },
+  "auth": {
+    "token_issuer": "myrobotaxi",
+    "token_audience": "telemetry"
+  }
+}
--- a/docs/simulator.md
+++ b/docs/simulator.md
@@ -11,6 +11,13 @@ The simulator requires mTLS certificates. Generate them once:
 ```bash
 mkdir -p certs
 
+# Homebrew openssl@3 ≥ 3.6 ships without /opt/homebrew/etc/openssl@3/openssl.cnf
+# and every `openssl req` call fails looking for it. Pointing OPENSSL_CONF at
+# /dev/null tells openssl to use compiled-in defaults, which is all this
+# recipe needs. macOS LibreSSL at /usr/bin/openssl isn't a drop-in
+# replacement because it lacks `-copy_extensions`.
+export OPENSSL_CONF=/dev/null
+
 # CA
 openssl ecparam -name prime256v1 -genkey -noout -out certs/ca.key
 openssl req -new -x509 -key certs/ca.key -out certs/ca.crt -days 365 -subj "/CN=MyRoboTaxi Dev CA"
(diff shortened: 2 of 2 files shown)


tnando:
**Claude finished @tnando's task in 1m 52s** —— [View job](https://github.com/tnando/my-robo-taxi-telemetry/actions/runs/24755925929)

---
### Claude PR Review — #185

## Review Summary

**Verdict: APPROVE**

**Domain focus:** none (general review only — docs/config only PR, no Go touched)

### Findings

#### Critical (must fix)
_None._

#### Warnings (should fix)
_None._

#### Suggestions
_None._

### Verification performed

**`configs/dev-notls.json`** (new, 38 lines)
- Valid JSON.
- Structure is identical to `configs/dev.json` in every nested block (`server`, `database`, `telemetry`, `drives`, `websocket`, `auth`) — only the three `tls` paths are empty strings. That matches exactly what the composition root expects:

  ```go
  // cmd/telemetry-server/main.go:236-240
  if cfg.TLS().CertFile == "" || cfg.TLS().KeyFile == "" {
      logger.Warn("TLS cert/key not configured — Tesla mTLS port will serve plain TCP (dev only)", ...)
  } else {
      teslaTLS, err := buildTeslaTLS(cfg.TLS())
      ...
  }
  ```

  So the empty-cert branch is explicitly a documented dev-only path with a runtime warning — no code changes needed.
- Referenced by `docs/ops-cli.md:176, 204, 208` (introduced in #184 but never committed). This PR fixes that dangling reference so the `--dev` ops-cli onboarding path actually works.
- No secrets in the committed file — DB URL / auth secret still come from env vars (per CLAUDE.md "Secrets via environment variables only"). ✅
- File is 38 lines, well under the 300-line cap. ✅

**`docs/simulator.md`** (+8 lines)
- The fix is correct: Homebrew `openssl@3` ≥ 3.6 does ship without `/opt/homebrew/etc/openssl@3/openssl.cnf`, and every `openssl req` call unconditionally opens `OPENSSL_CONF` on startup — pointing it at `/dev/null` forces compiled-in defaults, which is all this recipe actually needs (no sections are referenced in the `-subj` / `-addext` invocations).
- Comment block explains the *why* clearly, including the non-obvious LibreSSL caveat (`/usr/bin/openssl` lacks `-copy_extensions`) — future readers won't be tempted to "simplify" by dropping the env var.
- Scoping is correct: `export` at the top of the block, `unset` at the end. Users who paste the block into a long-lived shell won't leak `OPENSSL_CONF=/dev/null` into unrelated openssl work afterward.
- The underlying recipe (`-addext "subjectAltName=DNS:localhost,IP:127.0.0.1"` + `-copy_extensions copyall`) is unchanged, so the SAN-bearing server cert and VIN-CN client cert are preserved — matches the PR author's end-to-end verification.

### Domain-Specific Notes

No agent labels on this PR, so only the general gate applies. The CLAUDE.md rules that *could* apply to a config file — secrets handling, file size, location under `configs/` — all pass. Nothing in `internal/ws/`, `internal/store/`, the WebSocket contract, or the public SDK surface was touched, so `sdk-architect` / `contract-guard` are not triggered.

Clean, small, well-scoped docs+config unblock. Ship it.

---

- [x] Read PR context + CLAUDE.md
- [x] Fetch diff (`git diff origin/main...HEAD`)
- [x] Verify `dev-notls.json` matches composition root expectations (`main.go:236`)
- [x] Verify `dev-notls.json` structure vs `dev.json` (identical except empty TLS paths)
- [x] Verify `docs/simulator.md` openssl recipe + scoping of `OPENSSL_CONF`
- [x] Post verdict
 · `thomasnandola/myr-39-fix-docssimulatormd-openssl-recipe-for-homebrew-openssl3`

PR #191, opened 2026-04-25
MYR-43 Stop fetching navRouteCoordinates JSON on every WS broadcast

tnando:
## Summary

Fixes a Supabase egress regression where the WS broadcaster's VIN→ID lookup was selecting the full `Vehicle` row — including the heavy `navRouteCoordinates` JSON — on every Tesla telemetry frame, just to read `id` or `userId`. Over the 2026-04-01 → 2026-04-25 cycle this drove **6.87 GB / 5 GB** egress (137% of free tier) and put the org in grace until 2026-05-23.

## Root cause

`pg_stat_statements` showed one query at 663,851 calls (≈18/min, 24/7):

```sql
SELECT "id", "userId", "vin", "name", ..., "navRouteCoordinates", "lastUpdated"
FROM "Vehicle" WHERE "vin" = $1
```

Source: `vinResolverAdapter.GetByVIN` and `vehicleOwnerAdapter.GetVehicleOwner` in `cmd/telemetry-server/adapters.go` called `repo.GetByVIN` (24 columns) and read only `.ID` / `.UserID`. The existing `vinCache` in `internal/store/` was wired into the persistence writer but not the broadcaster.

## Fix

1. **Slim query**: `queryVehicleIDsByVIN` selects only `id, userId` (~50 B/row vs ~1-5 KB).
2. **Slim repo method**: `VehicleRepo.GetIDsByVIN(ctx, vin) (id, userID, err)`.
3. **Refactor `VINCache`**: now backed by `GetIDsByVIN`, stores `(vehicleID, userID)` per VIN, exposes `ResolveID` and `ResolveOwner` — both share the same cache entry. Type is exported so `cmd/telemetry-server` can construct one shared instance.
4. **Wire shared cache** into `vinResolverAdapter`, `vehicleOwnerAdapter` (used by `/api/vehicle-status/{vin}` and `/api/fleet-config/{vin}`), and the WS broadcaster path.

Both identifiers are immutable for the lifetime of a vehicle row, so cache entries live forever. `ErrVehicleNotFound` is also cached to prevent repeat lookups for unknown VINs; transient errors are not cached.

## Expected impact

- Per-call payload: ~1-5 KB → ~50 B (50× reduction) on cache miss.
- After warmup: zero egress for VIN→ID and VIN→owner resolution (sync.Map hit).
- Eliminates effectively all 6 GB of recurring egress traced to this query in `pg_stat_statements`.

## Contract impact

None. No WebSocket messages, REST shapes, DB columns, or public SDK API change. Pure backend perf fix touching `internal/store/`, `internal/ws/`-adjacent (`cmd/telemetry-server/adapters.go`), and `cmd/telemetry-server/main.go`.

## Test plan

- [x] `go vet ./...`
- [x] `go test ./... -short` — all packages pass (one pre-existing flake `TestReceiver_GracefulShutdown` passed 3/3 on retry, unrelated to this PR)
- [x] New unit tests in `internal/store/vin_cache_test.go` cover: cache hit/miss, owner-resolution reuses ID-cache entry, miss is cached, transient errors not cached
- [x] New integration test `TestVehicleRepo_GetIDsByVIN` covers existing-vehicle and missing-vehicle paths against the test pool
- [ ] After merge + deploy: confirm in `pg_stat_statements` that the 24-column `Vehicle WHERE vin = $1` query stops growing call count

## Out of scope

- Reducing `Drive.routePoints` writes (separate concern; current size is fine).
- Schema changes to the `Vehicle` table (Prisma-owned, write-only restricted).

### DIFF
--- a/.github/workflows/claude-review.yml
+++ b/.github/workflows/claude-review.yml
@@ -59,7 +59,13 @@ jobs:
           echo "Linked issue: ${ISSUE_NUM:-none}"
 
       - name: Run Claude review
-        uses: anthropics/claude-code-action@v1
+        # Pinned to v1.0.104. v1.0.105 (Apr 23 23:24) bumped the underlying
+        # Agent SDK to 0.2.119 and broke the action with an internal
+        # "directory mismatch for tsconfig.json" error that aborts Claude's
+        # session before it can post a review. Symptom: workflow exits
+        # successfully but the sticky comment is left at "I'll analyze this
+        # and get back to you." Bump cautiously and re-test on a real PR.
+        uses: anthropics/claude-code-action@0766301cba8671db92e3025984e2fd038ad48ff7 # v1.0.104
         with:
           claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
           github_token: ${{ secrets.GH_PAT }}
@@ -145,7 +151,13 @@ jobs:
           fetch-depth: 0
           token: ${{ secrets.GH_PAT }}
 
-      - uses: anthropics/claude-code-action@v1
+      # Pinned to v1.0.104. v1.0.105 (Apr 23 23:24) bumped the underlying
+      # Agent SDK to 0.2.119 and broke the action with an internal
+      # "directory mismatch for tsconfig.json" error that aborts Claude's
+      # session before it can post a review. Symptom: workflow exits
+      # successfully but the sticky comment is left at "I'll analyze this
+      # and get back to you." Bump cautiously and re-test on a real PR.
+      - uses: anthropics/claude-code-action@0766301cba8671db92e3025984e2fd038ad48ff7 # v1.0.104
         with:
           claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
           github_token: ${{ secrets.GH_PAT }}
--- a/cmd/telemetry-server/adapters.go
+++ b/cmd/telemetry-server/adapters.go
@@ -62,32 +62,37 @@ func resolveDebugFieldsGate(devMode bool, token string) (debugFieldsGate, error)
 	}
 }
 
-// vinResolverAdapter adapts store.VehicleRepo (returns Vehicle) to the
-// ws.VINResolver interface (returns vehicleID string).
+// vinResolverAdapter adapts store.VINCache to the ws.VINResolver interface
+// (returns vehicleID string). Backing the WS broadcaster path with the cache
+// avoids fetching the full Vehicle row (including the heavy navRouteCoordinates
+// JSON) on every telemetry frame — the slim two-column query runs once per VIN
+// for the lifetime of the process.
 type vinResolverAdapter struct {
-	repo *store.VehicleRepo
+	cache *store.VINCache
 }
 
 func (a *vinResolverAdapter) GetByVIN(ctx context.Context, vin string) (string, error) {
-	v, err := a.repo.GetByVIN(ctx, vin)
+	id, err := a.cache.ResolveID(ctx, vin)
 	if err != nil {
 		return "", fmt.Errorf("resolve VIN: %w", err)
 	}
-	return v.ID, nil
+	return id, nil
 }
 
-// vehicleOwnerAdapter adapts store.VehicleRepo to the
-// telemetry.VehicleOwnerLookup interface (returns owning user ID).
+// vehicleOwnerAdapter adapts store.VINCache to the
+// telemetry.VehicleOwnerLookup interface (returns owning user ID). Shares
+// the same cache instance as vinResolverAdapter, so a single DB lookup per
+// VIN serves both ID and owner resolution for the lifetime of the process.
 type vehicleOwnerAdapter struct {
-	repo *store.VehicleRepo
+	cache *store.VINCache
 }
 
 func (a *vehicleOwnerAdapter) GetVehicleOwner(ctx context.Context, vin string) (string, error) {
-	v, err := a.repo.GetByVIN(ctx, vin)
+	userID, err := a.cache.ResolveOwner(ctx, vin)
 	if err != nil {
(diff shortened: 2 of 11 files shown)


tnando:
**Claude finished @tnando's task in 2m 10s** —— [View job](https://github.com/tnando/my-robo-taxi-telemetry/actions/runs/24938196847)

---
I'll analyze this and get back to you.

tnando:
**Claude finished @tnando's task in 3m 3s** —— [View job](https://github.com/tnando/my-robo-taxi-telemetry/actions/runs/24938196847)

---
I'll analyze this and get back to you.

tnando:
**Claude finished @tnando's task in 1m 44s** —— [View job](https://github.com/tnando/my-robo-taxi-telemetry/actions/runs/24938485924)

---
### PR Review

- [x] Read PR diff
- [x] Review changed files
- [x] Apply general Go review checks
- [x] Post final verdict

---

## Review Summary

**Verdict: APPROVE**

**Domain focus:** none specified — applied general Go + project rule review. Changes touch `internal/store/`, `cmd/telemetry-server/`, and `.github/workflows/` (action pin). No WebSocket/REST/SDK contract changes (PR confirms "no contract impact"), so contract-guard and sdk-architect gates are not implicated by these specific edits.

This is a well-scoped, surgical fix for a real production problem (137% of free-tier egress traced to one over-fetching query). The refactor is clean, tests are good, and the diff is appropriately narrow.

### Findings

#### Critical (must fix)
None.

#### Warnings (should fix)
None blocking. One thing to verify (see Suggestions).

#### Suggestions

1. **Forever-cached `userId` rests on an immutability assumption worth double-checking against the Prisma schema.**
   `internal/store/vin_cache.go:30-33` documents that both `id` and `userId` are "immutable for the lifetime of a vehicle row." `id` is genuinely immutable (cuid PK). `userId` is only immutable if the Prisma side never reassigns `Vehicle.userId` — i.e., no ownership transfer / re-link flow. If a user ever transfers a vehicle to another account in the Next.js app, this long-lived process will keep serving the old owner from `ResolveOwner`, which then drives `/api/vehicle-status/{vin}`, `/api/fleet-config/{vin}`, and `VehicleOwnerLookup`-gated paths. Worth confirming with `my-robo-taxi`'s Prisma schema before assuming forever-cache is safe. If reassignment is possible (even rarely), a TTL or invalidation hook would be safer than indefinite caching.

2. **Thundering herd on cold start for concurrent VINs (minor).**
   `resolve` is not deduplicated — N concurrent goroutines hitting the same uncached VIN will each issue a DB query before the first store wins. With ~5K vehicles connecting at roughly the same moment after a deploy, this is a one-off burst, not a recurring cost. If it becomes an issue, `golang.org/x/sync/singleflight` slots in cleanly here. Not a blocker.

3. **Test struct literal is a bit noisy — consider a tiny named type.**
   `internal/store/writer_test.go` and `vin_cache_test.go` repeat `map[string]struct{ id, userID string }{...}` ~10 times. A small package-private alias like `type idPair = struct{ id, userID string }` would make the table-driven tests easier to read. Pure cosmetic.

### Domain-Specific Notes (general checks passed)

- **Error handling:** All paths wrap with `fmt.Errorf("...: %w", err)` and propagate `ErrVehicleNotFound` via `errors.Is`. Transient errors are explicitly *not* cached — the comment at `vin_cache.go:95` calls this out. Good.
- **Context propagation:** `GetIDsByVIN` and both `Resolve*` methods take `context.Context` first. ✓
- **Logging hygiene:** VINs go through `redactVIN` everywhere. No full-VIN leaks. ✓ (CLAUDE.md non-negotiable.)
- **No new globals, no `init()`, no `panic()`.** ✓
- **Interface at consumer site:** `vinIDLookup` is defined in `internal/store/vin_cache.go` next to its consumer, and `VehicleRepo` happens to satisfy it. Matches the project convention. ✓
- **File sizes:** `vin_cache.go` is 107 lines, `vehicle_repo.go` ~57 lines added, well under the 300-line cap. ✓
- **Dependency direction:** `cmd/telemetry-server` → `internal/store`. ✓
- **`sync.Map` usage:** Used correctly here — read-heavy (one write per VIN, many reads forever after). ✓
- **Concurrency safety:** `sync.Map.Load`/`Store` are race-safe; the `vinIDs` struct is comparable (two strings) so the `== vinMissSentinel` check works as intended. ✓
- **Tests:**
  - Table-driven with `t.Run` subtests. ✓
  - `TestVINCache_ResolveOwnerReusesSameEntry` directly verifies the PR's central claim that one entry serves both methods. ✓
  - `TestVINCache_MissCachedPreventsRepeatedLookup` now mixes `ResolveID` then `ResolveOwner` to prove the negative cache is shared too — nice improvement.
  - `TestVehicleRepo_GetIDsByVIN` covers both the hit and `ErrVehicleNotFound` paths against the real test pool.
  - One nit: in `TestVehicleRepo_GetIDsByVIN`, the comment "seedVehicle hardcodes this owner" is brittle — if the seed helper ever changes the hardcoded `'user_001'`, this test will fail mysteriously. A query like `SELECT "userId" FROM "Vehicle" WHERE id = $1` to discover the expected value would be sturdier, but this is minor.
- **Adapter rename / field plumbing:** `vinResolverAdapter.repo → cache` and `vehicleOwnerAdapter.repo → cache` correctly thread the *same* `*VINCache` instance from `main.go:166` through both adapters and the fleet-config endpoint, which is what makes ID and owner lookups share one entry. ✓
- **Workflow pin to SHA `0766301`:** Clearly justified by the in-file comment about v1.0.105 breaking the action. Pinning by SHA is good supply-chain hygiene — though this also bypasses workflow-level Dependabot updates, so adding a renovate/dependabot annotation later is worth considering.

### Net

Clean, focused fix with appropriate test coverage. The only thing I'd verify before merge is the `userId` immutability assumption against the Prisma schema in `my-robo-taxi/` — if vehicle ownership can be reassigned, the cache needs invalidation. If reassignment is impossible by design, this PR is good as-is.

PR #198, opened 2026-04-30
MYR-14 Atomic field-group declaration + accumulator generalization

tnando:
## Summary

- Declares the 4 v1 atomic field groups (`navigation`, `charge`, `gps`, `gear`) as a single source of truth in `internal/ws/atomic_groups.go`; `groupOf()` becomes the partition predicate used by the broadcaster, and `isNavField` now derives from this map.
- Generalizes the navigation-only accumulator into a per-group `groupAccumulator` keyed by `(atomicGroupID, VIN)`. In v1 only `groupNavigation` registers a slot; charge/gps/gear pass through the immediate-broadcast path because their atomicity is upstream-guaranteed (Tesla 500 ms bucket / lat-lng co-emission / synchronous status derivation).
- Updates `vehicle-state-schema.md` §2.1–§2.4 + change log and `websocket-protocol.md` §3.2 / §3.2.1 / §4.1.4 / §8 / DV-01 in the same PR per contract-guard.

## Decisions recorded in the docs

- **500 ms window kept** (not the 200 ms in the original AC). Tesla's vehicle-side emission bucket floor is 500 ms; a shorter server window cannot improve atomicity because Tesla itself cannot deliver companion fields any faster. NFR-3.2 amendment in `requirements.md` remains pending under DV-01.
- **No accumulator for charge/gps/gear** — each has a documented upstream guarantee (`vehicle-state-schema.md` §2.2 / §2.3 / §2.4). The accumulator is generalized so a future MYR can opt any group in without a refactor.
- **Snapshot-on-connect unchanged.** REST `/snapshot` reads from `store.Vehicle` which is already group-consistent; no WS replay path was modified.

## Known gap (out of scope; tracked separately)

`status` is not recomputed on speed-only telemetry frames. A vehicle going from `parked, 0 mph` to `speed > 0` with no fresh `gear` event stays at `status: parked` until the next gear update. Recomputing on speed alone would emit an unsynchronized `status`/`gearPosition` pair — a worse atomicity violation. Documented in `vehicle-state-schema.md` §2.4 and cross-linked from §1.1's `status` row. A follow-on Linear issue with `relatedTo MYR-14` will be filed; ux-audit confirmed the gap is user-visible in `../my-robo-taxi/src/features/vehicles/components/SharedViewerScreen.tsx:42`.

## Pre-PR gates

- `go vet ./...` — clean
- `golangci-lint run ./...` — 0 issues
- `go test ./...` — green (incl. new `TestGroupOf`, `TestAtomicGroupMembersDisjoint`, `TestGroupAccumulator_DifferentGroupsIndependent`, plus the existing nav-accumulator suite renamed and updated for the new API; `TestAtomicGroupCompleteness` still PASS)
- `contract-guard` — PASS
- `ux-audit` — PASS (cross-link from §1.1 to §2.4 applied per audit suggestion)
- `sdk-architect` final review — APPROVED

## Test plan

- [x] `go test ./internal/ws/...` and `go test ./tests/contract/...` pass locally
- [x] `golangci-lint run ./...` clean
- [x] No wire bytes change — manual trace through `handleTelemetry` confirms identical JSON shape and same drive-end / disconnect drain semantics
- [x] Cross-group isolation test (`TestGroupAccumulator_DifferentGroupsIndependent`) confirms groups never leak fields into each other's batches


tnando:
**Claude finished @tnando's task in 2m 54s** —— [View job](https://github.com/tnando/my-robo-taxi-telemetry/actions/runs/25146823438)

---
### Claude PR Review — MYR-14

- [x] Read PR diff
- [x] Review new files (`atomic_groups.go`, `group_accumulator.go`)
- [x] Review modified files (`broadcaster.go`, `nav_broadcast.go`, `field_mapping.go`)
- [x] Review tests
- [x] Post verdict

---

## Review Summary

**Verdict: APPROVE**

**Domain focus:** No `agent:*` labels were attached, so general Go review only (architecture, concurrency, tests, contract adherence touch‑points). PR description indicates `sdk-architect`, `contract-guard`, and `ux-audit` already passed.

This is a well‑executed refactor: a single source of truth (`atomicGroupMembers`) replaces a navigation‑specific helper, and the accumulator is now keyed by `(group, vin)` so future groups (e.g., MYR‑40 charge if the upstream guarantee changes) can opt in without a structural change. Public behaviour is unchanged for v1 because only `groupNavigation` registers a slot.

### Findings

#### Critical (must fix)
None.

#### Warnings (should fix)
None.

#### Suggestions (non‑blocking)

1. **`flushGroup` synthesises `lastUpdated` at flush time** — `internal/ws/nav_broadcast.go:103` uses `time.Now().UTC()` for the wire `lastUpdated`. The original `payload.CreatedAt` of the most‑recent contributing telemetry frame is more accurate (and is what the immediate path uses on `nav_broadcast.go:65`). For a 500 ms window this is a sub‑second skew — not a contract violation — but threading the latest `CreatedAt` through `groupAccumulator.Add` (e.g. `lastSeen time.Time`) would keep timestamps consistent across the two delivery paths and avoid client‑side ordering surprises if the system clock skews.

2. **`groupAccumulator.Stop` has a small in‑flight window** — `internal/ws/group_accumulator.go:140-151`. `time.Timer.Stop()` does not synchronise with an `AfterFunc` goroutine that has already started, so a `timerFired` already past `a.mu.Unlock()` could call `onFlush` after `Stop` returns. The current implementation is *mostly* safe because `Stop` drains `pending` under the lock, so a `timerFired` that wins the lock race after `Stop` will read an empty map and skip the callback — but a `timerFired` that acquired the lock *before* `Stop` will fire its callback. In a real shutdown that crosses through `hub.Stop` this could touch a stopped hub. A `sync.WaitGroup` (or a `done` channel checked inside `timerFired`) would close the gap if you ever need strict shutdown ordering. Not blocking for v1 — the test cleanup order (`b.Stop` before `hub.Stop`) keeps the hub alive during the in‑flight window.

3. **`groupOf` does an O(N) map‑of‑maps scan per call** — `internal/ws/atomic_groups.go:66-73`. With 4 groups × ~13 fields total, perf is fine, but every telemetry frame iterates the outer map in random order. Inverting at package init into a flat `map[string]atomicGroupID` would be O(1) and remove the dependence on the disjointness invariant for correctness (you'd still want `TestAtomicGroupMembersDisjoint` as a guardrail at construction time). Pure micro‑optimisation; ignore unless a profile flags it.

4. **`TestIsNavField` is now redundant with `TestGroupOf`** — `internal/ws/group_accumulator_test.go:11-39`. Since `isNavField` is a one‑line wrapper around `groupOf`, the two tests cover the same logic. Consider deleting `TestIsNavField` or replacing it with one row that verifies the wrapper semantics (`groupOf returns groupGPS → isNavField false`).

5. **No Go‑side cross‑check against schema `x-atomic-groups`** — `atomicGroupMembers` (internal field names) and `vehicle-state.schema.json#x-atomic-groups` (wire names) are *both* sources of truth that must stay aligned. `tests/contract/fixtures_test.go:352` (`TestAtomicGroupCompleteness`) checks fixture↔schema, and `TestAtomicGroupMembersDisjoint` checks Go internal consistency, but nothing translates `atomicGroupMembers` through `internalToClientField` and asserts equality with `x-atomic-groups`. Drift between the two would be silent. Pre‑existing structural gap, not introduced by this PR — worth a follow‑up issue if you want belt‑and‑braces.

### Domain‑Specific Notes

- **Contract docs updated in‑PR** (`vehicle-state-schema.md` §2.1–2.4 + change log; `websocket-protocol.md` §3.2 / §3.2.1 / §4.1.4 / §8 / DV‑01) — meets the contract‑guard requirement from CLAUDE.md "Contracts" section.
- **500 ms window justification** is now anchored in §2.2 ("Tesla's vehicle‑side emission bucket floor") and the package doc comment on `defaultGroupFlushInterval` — good, this is exactly the kind of "why" comment that survives the next refactor.
- **`status` derivation gap** is documented in‑PR (vehicle‑state‑schema.md §2.4) and cross‑linked from §1.1 per the ux‑audit suggestion. Following the architect's principle of "one observable invariant per change", deferring the speed‑only recompute to a separate Linear issue is the right call — a partial fix here would introduce a worse atomicity violation between `status` and `gearPosition`.
- **Drive‑end / disconnect drain semantics preserved** — `handleDriveEnded` calls `b.groups.Flush(groupNavigation, vin)` (which cancels timer + clears in one shot, so the prior `Clear` call is correctly omitted), and `handleConnectivity` clears on disconnect. Wire behaviour is byte‑identical to pre‑refactor for v1.
- **All files comfortably under 300 lines** (largest non‑test: `broadcaster.go` at 260).

### Wire compatibility check
- `vehicle_update` shape unchanged — same `mapFieldsForClient` → `BroadcastMasked` path on both the immediate and accumulated branches.
- No new message types, no field renames.
- Frontend `useVehicleStream` consumers unaffected.
 · branch: `thomasnandola/myr-14-implement-atomic-field-group-debouncing-navchargegpsgear`

PR #199, opened 2026-05-01
MYR-16 AES-256-GCM column encryption foundation

tnando:
## Summary

- Adds the `internal/cryptox` package: AES-256-GCM primitives, versioned ciphertext, `KeySet` rotation, public `Encryptor` interface.
- Wires `cryptox.LoadKeySetFromEnv` + `NewEncryptor` at startup so the binary fails-fast on missing/invalid `ENCRYPTION_KEY`.
- Adds the `docs/contracts/key-rotation.md` contract doc and updates `data-classification.md` §3.3 + the contracts README index.

## Why foundation-only

Every P1 column requiring encryption per `data-classification.md` §3 lives in the **Prisma-owned schema in `../my-robo-taxi/`** (Account.access_token / refresh_token / id_token, Vehicle GPS columns, Drive.routePoints). Per CLAUDE.md the telemetry server cannot modify Prisma-owned tables; the actual column flips need cross-repo Prisma migrations + symmetric encrypt/decrypt on both sides + backfill + plaintext drop. That's a 3+ PR sequence across 2 repos. This PR ships the single-repo foundation; per-column rollouts will land as follow-on Linear issues (one per migration unit: OAuth tokens, Vehicle GPS coords, route blobs, plus a key-rotation production drill).

**No P1 column is yet encrypted on disk** — the `Encryptor` is constructed at startup but discarded with `_ = encryptor` per the architect's design memo (no dead constructor params).

## Decisions recorded in the contract docs

- **500 ms window kept** is a separate, unrelated decision in MYR-14. This PR's rotation contract is in `key-rotation.md`.
- **Ciphertext format:** `[1B version][12B nonce][N B ct + 16B GCM tag]`, base64-encoded for `Text` storage. Version `0x01` only emitted; `0x00` reserved as `ErrInvalidVersion`.
- **KeySet rotation:** active write version + 0..N retired-but-readable versions; decrypt fans out by version byte at position 0. Two env-var shapes (`ENCRYPTION_KEY` shorthand, or `ENCRYPTION_KEY_V{N}` + `ENCRYPTION_WRITE_VERSION`); both shapes set is an error.
- **Defense-in-depth:** `KeySet.String()` and `keySetEncryptor.LogValue()` both redact, so `slog.Any(...)` accidents never leak key material.

## Operational impact

- **`ENCRYPTION_KEY` is now a required env var** at server startup. CI excludes `/cmd/` from tests, so test gates aren't affected. `docker-compose.test.yml`, `.env.example`, and `fly.toml` all updated to document the new requirement.
- **Do NOT execute the rotation procedure in production yet** — `cryptox_decrypt_total{version=N}` is not yet wired (lands with the first column rollout PR). Step 6 of the rotation procedure cannot be safely completed without it. Status callout added at the top of `key-rotation.md`.

## Pre-PR gates

- `go vet ./...` — clean
- `golangci-lint run ./...` — 0 issues
- `go test ./...` — green (incl. new `TestEncryptor_*`, `TestLoadKeySetFromEnv_*`, `TestEncryptGCM_*`, `FuzzDecryptString` seed corpus)
- `security` review — PASS with two must-fixes applied (empty-string sentinel doc + explicit empty-`ENCRYPTION_WRITE_VERSION` test) and three defense-in-depth additions (`LogValue` redaction, fuzz seed corpus, nonce-space rotation guidance)
- `contract-guard` — PASS (after fixing one stale "eight contract documents" heading)
- `ux-audit` — PASS with three deployment-config gaps closed (`docker-compose.test.yml`, `.env.example`, `fly.toml`) + two doc clarifications
- `sdk-architect` final review — APPROVED

## Test plan

- [x] `go test ./internal/cryptox/...` — round-trip, nonce uniqueness (N=1000), tampered ct/tag, wrong key, short input, invalid key sizes, version-byte routing across V1/V2, unknown-version error path, KeySet redaction, env-var error matrix
- [x] `go test ./...` — full suite green; no other package affected
- [x] Manual: confirmed binary refuses to start without `ENCRYPTION_KEY` (fail-fast)
- [ ] Manual smoke before merge: deploy to staging with a real `ENCRYPTION_KEY` and confirm `encryptor initialized write_version=1` appears in logs

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -12,6 +12,13 @@ DATABASE_URL=postgres://postgres:dev@postgres:5432/myrobotaxi
 # Generate with: openssl rand -hex 32
 AUTH_SECRET=CHANGEME-run-openssl-rand-hex-32
 
+# ---- Encryption (NFR-3.23, NFR-3.24) -----------------------------------------
+# AES-256 key for column-level encryption. Required at startup.
+# Generate with: openssl rand -base64 32
+# For key rotation, switch to the versioned shape (ENCRYPTION_KEY_V1,
+# ENCRYPTION_KEY_V2, ENCRYPTION_WRITE_VERSION) — see docs/contracts/key-rotation.md.
+ENCRYPTION_KEY=CHANGEME-run-openssl-rand-base64-32
+
 # ---- Tesla HTTP Proxy --------------------------------------------------------
 # Path to the Tesla app private key used to sign Fleet API commands.
 # The proxy is only started when this is set and the file exists.
--- a/cmd/telemetry-server/main.go
+++ b/cmd/telemetry-server/main.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/tnando/my-robo-taxi-telemetry/internal/auth"
 	"github.com/tnando/my-robo-taxi-telemetry/internal/config"
+	"github.com/tnando/my-robo-taxi-telemetry/internal/cryptox"
 	"github.com/tnando/my-robo-taxi-telemetry/internal/drives"
 	"github.com/tnando/my-robo-taxi-telemetry/internal/events"
 	"github.com/tnando/my-robo-taxi-telemetry/internal/geocode"
@@ -97,6 +98,27 @@ func run() error { //nolint:funlen // composition root — sequential dependency
 	reg := prometheus.NewRegistry()
 	reg.MustRegister(collectors.NewGoCollector())
 	reg.MustRegister(collectors.NewProcessCollector(collectors.ProcessCollectorOpts{}))
+
+	// --- Column encryption foundation (NFR-3.23, NFR-3.24) ---
+	// Loads the AES-256-GCM key set so the binary fails-fast on missing
+	// or invalid ENCRYPTION_KEY at startup. The Encryptor is constructed
+	// but not yet wired into the store layer; per-table column rollouts
+	// are tracked by follow-on Linear issues that require coordinated
+	// Prisma schema changes in ../my-robo-taxi/. See
+	// docs/contracts/key-rotation.md and docs/contracts/data-classification.md
+	// §3.3 for the encryption contract.
+	keySet, err := cryptox.LoadKeySetFromEnv()
+	if err != nil {
+		return fmt.Errorf("loading encryption key set: %w", err)
+	}
+	encryptor, err := cryptox.NewEncryptor(keySet)
+	if err != nil {
+		return fmt.Errorf("constructing encryptor: %w", err)
+	}
+	_ = encryptor // foundation: column wiring lands in follow-on PRs (MYR-16 cross-repo rollout issues)
+	logger.Info("encryptor initialized",
+		slog.Int("write_version", int(keySet.WriteVersion())),
+	)
 
 	// --- Database connection ---
 	db, err := store.NewDB(ctx, cfg.Database(), logger.With(slog.String("component", "store")), store.NoopMetrics{})
--- a/cmd/telemetry-server/main.go
+++ b/cmd/telemetry-server/main.go
@@ -45,7 +45,7 @@ func main() {
 	}
 }
 
-func run() error { //nolint:funlen // composition root — sequential dependency wiring
+func run() error { //nolint:funlen,cyclop // composition root — sequential dependency wiring; each new dependency adds a guard, not branching logic
 	// --- Flag parsing ---
 	var (
 		configPath = flag.String("config", "", "path to JSON configuration file")
--- a/cmd/telemetry-server/main.go
+++ b/cmd/telemetry-server/main.go
@@ -6,8 +6,6 @@ package main
 
 import (
 	"context"
-	"crypto/tls"
-	"crypto/x509"
 	"flag"
 	"fmt"
 	"log/slog"
@@ -19,9 +17,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/collectors"
 
-	"github.com/tnando/my-robo-taxi-telemetry/internal/auth"
 	"github.com/tnando/my-robo-taxi-telemetry/internal/config"
-	"github.com/tnando/my-robo-taxi-telemetry/internal/cryptox"
 	"github.com/tnando/my-robo-taxi-telemetry/internal/drives"
 	"github.com/tnando/my-robo-taxi-telemetry/internal/events"
 	"github.com/tnando/my-robo-taxi-telemetry/internal/geocode"
@@ -45,7 +41,7 @@ func main() {
 	}
 }
 
-func run() error { //nolint:funlen,cyclop // composition root — sequential dependency wiring; each new dependency adds a guard, not branching logic
+func run() error { //nolint:funlen // composition root — sequential dependency wiring; helpers extracted to wiring.go
 	// --- Flag parsing ---
 	var (
 		configPath = flag.String("config", "", "path to JSON configuration file")
@@ -100,25 +96,11 @@ func run() error { //nolint:funlen,cyclop // composition root — sequential dep
 	reg.MustRegister(collectors.NewProcessCollector(collectors.ProcessCollectorOpts{}))
 
 	// --- Column encryption foundation (NFR-3.23, NFR-3.24) ---
-	// Loads the AES-256-GCM key set so the binary fails-fast on missing
-	// or invalid ENCRYPTION_KEY at startup. The Encryptor is constructed
-	// but not yet wired into the store layer; per-table column rollouts
-	// are tracked by follow-on Linear issues that require coordinated
-	// Prisma schema changes in ../my-robo-taxi/. See
-	// docs/contracts/key-rotation.md and docs/contracts/data-classification.md
-	// §3.3 for the encryption contract.
-	keySet, err := cryptox.LoadKeySetFromEnv()
+	encryptor, err := setupEncryption(logger)
 	if err != nil {
-		return fmt.Errorf("loading encryption key set: %w", err)
-	}
-	encryptor, err := cryptox.NewEncryptor(keySet)
-	if err != nil {
-		return fmt.Errorf("constructing encryptor: %w", err)
+		return err
 	}
 	_ = encryptor // foundation: column wiring lands in follow-on PRs (MYR-16 cross-repo rollout issues)
-	logger.Info("encryptor initialized",
-		slog.Int("write_version", int(keySet.WriteVersion())),
-	)
 
 	// --- Database connection ---
 	db, err := store.NewDB(ctx, cfg.Database(), logger.With(slog.String("component", "store")), store.NoopMetrics{})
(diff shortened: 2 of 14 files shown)


tnando:
**Claude finished @tnando's task in 2m 24s** —— [View job](https://github.com/tnando/my-robo-taxi-telemetry/actions/runs/25234808929)

---
## Review Summary

**Verdict: APPROVE**

**Domain focus:** General Go review (no `agent:*` labels on this PR — applying base-rate review per `pr-reviewer.md`).

This is a clean, well-tested, well-documented foundation PR. The "land the primitive but don't yet wire it into a column" approach is the right call given that every P1 column lives in the Prisma-owned schema in `../my-robo-taxi/` and per-column rollouts require coordinated migrations across two repos.

### Findings

#### Critical (must fix)
- None.

#### Warnings (should fix)
- None blocking. See suggestions below.

1. **`cmd/telemetry-server/main.go` is now 411 lines** (was 388 pre-PR; +23 from this PR). CLAUDE.md caps non-test files at 300 lines. This is a pre-existing violation worsened by this PR rather than introduced by it, and the existing `//nolint:funlen,cyclop` on `run()` is appropriate for a composition root. Worth a follow-on to break out helpers like `wireEncryption(ctx, logger)`, `wireWriter(...)`, `wireFleetConfig(...)` so each new module add stops compounding the file-size drift. Not blocking.

2. **`_ = encryptor // foundation: column wiring lands in follow-on PRs`** (`cmd/telemetry-server/main.go:118`). The discard is intentional and the comment explains it well. One alternative worth considering for the next PR: stash the `Encryptor` on a top-level `deps` struct now (even unused) so the column-rollout PRs are pure additions to the wiring graph rather than re-introducing a discarded variable. Optional — current shape is fine for a foundation-only PR.

3. **Empty-string sentinel design** (`encryptor.go:18-28`). The "empty in → empty out" semantic is explicitly documented as "absent value, not encrypted empty payload". This is the correct call for nullable Postgres `Text` columns, and the doc comment forecloses the footgun. Worth adding a runtime assertion in the future column-write path (`if col == "" && !col.Nullable { return errExplicitlyEmptyForbidden }`) so the contract is enforced at the call site, not just documented.

4. **Nonce-space rotation guidance** (`key-rotation.md:33`) is well-reasoned (~1B encryptions per key as the practical cap, well below NIST's 2³² bound). When the GPS rollout lands, consider exposing `cryptox_encrypt_total{version="N"}` as a sibling counter to `cryptox_decrypt_total` so operators can dashboard-track approach to the rotation cap rather than rederiving it from per-column write rates. Out of scope for this PR.

5. **Test for `decodeAndValidate` URL-safe vs standard ambiguity.** `TestDecodeAndValidate_URLSafeBase64` covers the URL-safe path. Consider adding a test that confirms a string that is *valid* under URL-safe encoding but *invalid* under std (or vice versa) routes correctly — there exist 32-byte payloads whose std and URL-safe encodings differ on `+/-` and `//_`, so the fallback path on line 182 is non-obvious. The current test seeds 0xFF bytes which produces a base64 with no special chars (all `/` or `+` being relevant). Minor — doesn't change correctness.

6. **`FuzzDecryptString` uses `&testing.T{}`** (`encryptor_test.go:237`). Constructing a zero-value `*testing.T` to thread into the helper works in practice but is brittle (helpers calling `t.Helper()` or `t.Setenv()` would NPE). Consider lifting the seed-corpus encryptor construction inline instead of via `newTestKeySet`. Cosmetic.

### Domain-Specific Notes

#### Security (defense-in-depth is excellent)
- `KeySet.String()` and `keySetEncryptor.LogValue()` both redact — closes the `slog.Any("ks", ks)` accident vector.
- Reserved version `0x00 == invalid` defends against zero-init buffers being silently accepted on decrypt — good.
- Empty env vars treated as not-set prevents the "deployed with empty key" silent-failure mode.
- Mutual exclusivity check on `ENCRYPTION_KEY` vs `ENCRYPTION_KEY_V*` prevents config ambiguity.
- Authentication failure on tamper / wrong-key / wrong-version-key is unified by GCM design — wrap message stays generic per `aes_gcm.go:108-111`.
- `decodeAndValidate` accepts both std and URL-safe base64 — pragmatic tolerance for secret-store conventions without weakening validation.
- Fresh random nonce per call (`io.ReadFull(rand.Reader, ...)`) — correct GCM usage, nonce reuse would be catastrophic.
- TLS 1.2+ for transit (already present in `buildTeslaTLS`); at-rest encryption is the defense-in-depth complement that this PR establishes.

#### Tesla telemetry / data classification
- `data-classification.md §3.3` correctly anchors the foundation status. The 11 columns enumerated (Account.access_token/refresh_token/id_token, Vehicle GPS quartet + dest pair + origin pair + navRouteCoordinates, Drive.routePoints) match `key-rotation.md`'s rollout scope.
- `key-rotation.md`'s "do not execute the rotation procedure in production yet" callout is a strong operational guard — step 6 (zero-decrypt-window confirmation) is genuinely unsafe without `cryptox_decrypt_total`.

#### Cross-repo symmetry concern (out of scope, but worth flagging)
The Prisma side in `../my-robo-taxi/` will need to implement the *same* ciphertext format (`[1B version][12B nonce][N B ct + 16B tag]`, base64) to encrypt/decrypt symmetrically. `key-rotation.md §"Ciphertext format"` is the authoritative spec for this — recommend the first column-rollout Linear issue explicitly references the Node.js side reading this exact section as the implementation contract, so the two implementations don't diverge on padding/encoding choices.

#### Tests
- Round-trip across all interesting plaintext shapes ✓
- Nonce uniqueness (N=1000) ✓
- Tampered ct + tampered tag + wrong key ✓
- Cross-version decrypt + unknown version ✓
- Invalid version byte (0x00) ✓
- Env-var error matrix (8 cases) ✓
- Redaction smoke test ✓
- Fuzz seed corpus on `DecryptString` ✓
- All helpers properly use `t.Helper()` ✓

### Pre-PR gate verification (per PR description)
- `go vet ./...` — described as clean ✓
- `golangci-lint run ./...` — described as 0 issues ✓
- security review — applied, with two must-fixes incorporated ✓
- contract-guard — passing ✓
- ux-audit — passing with deployment configs (`docker-compose.test.yml`, `.env.example`, `fly.toml`) updated ✓

---

### Todo
- [x] Read CLAUDE.md and contract docs
- [x] Review `internal/cryptox/` package (aes_gcm, encryptor, key)
- [x] Review `cmd/telemetry-server/main.go` wiring
- [x] Review test coverage
- [x] Review contract docs (key-rotation.md, data-classification.md)
- [x] Post final verdict

Strong, foundation-only PR. Cryptography primitives are correctly used, tests cover the failure surface, contract docs are thorough, and the operational caveats (do-not-rotate-yet, no-column-encrypted-yet) are surfaced where operators will see them. Approve.
 | [Branch](https://github.com/tnando/my-robo-taxi-telemetry/tree/thomasnandola/myr-16-implement-aes-256-gcm-column-encryption-for-p1-fields)

tnando:
**Claude finished @tnando's task in 1m 35s** —— [View job](https://github.com/tnando/my-robo-taxi-telemetry/actions/runs/25283587639)

---
### Review Summary

**Verdict: APPROVE**

**Domain focus:** General Go review (no `agent:*` labels on this PR).

This re-review covers the new commit `0376505` since the prior approval. The cryptox package, contract docs, and startup wiring are unchanged from the last approved revision; the new commit is a pure code-organization refactor that extracts wiring helpers from `main.go` to `cmd/telemetry-server/wiring.go` to fit the CLAUDE.md 300-line cap.

### Findings

#### Critical (must fix)
- None.

#### Warnings (should fix)
- None.

1. **File-size compliance restored.** `main.go` is now **248 lines** (was 411), `wiring.go` is **241 lines**, `adapters.go` stays at **174 lines**. All cmd/telemetry-server files comfortably under the 300-line cap. This directly addresses suggestion #1 from the prior review.

2. **Extraction strategy is the right shape.** The header comment on `wiring.go:1-3` makes the intent explicit:
   > "None of these add abstraction over what `run()` already did inline — they are pure code-organization extractions."
   The helpers (`setupEncryption`, `setupAuthenticator`, `setupHTTPHandlers`, `setupTeslaTLS`, `setupFleetConfigEndpoint`, `buildTeslaTLS`) each map 1:1 to a previously-inline section of `run()`. No premature interfaces, no leaky abstractions. ✓

3. **`httpRouteDeps` parameter struct** (`wiring.go:68-80`) is a clean fix for `setupHTTPHandlers` having 11 dependencies. Field-name keying at the call site (`main.go:193-205`) is more readable than positional args and makes adding a new dep a localized edit. ✓

4. **`run()` `//nolint` comment updated** (`main.go:44`) to reference the helper extraction:
   > `//nolint:funlen // composition root — sequential dependency wiring; helpers extracted to wiring.go`
   Good — keeps the suppression honest about what was done.

5. **`_ = encryptor` discard** (`main.go:103`) preserved with the same accurate comment ("foundation: column wiring lands in follow-on PRs"). The previous review's suggestion #2 (stash on a `deps` struct) remains optional and is best deferred to the first column-rollout PR rather than churning this foundation again.

6. **Minor — `wiring.go` has no dedicated test file**, but each extracted helper is exercised transitively by integration tests via `run()` and the existing `adapters_test.go` covers the helpers that needed unit tests (`resolveDebugFieldsGate`, `proxyHTTPClient`). The pure pass-through wrappers in `wiring.go` aren't worth dedicated unit tests. Acceptable.

### Domain-Specific Notes

#### Cryptography (unchanged since prior approval)
The cryptox package has not been touched since commit `9efc2d2` — last review's security analysis still stands:
- Versioned ciphertext format `[1B version][12B nonce][N B ct + 16B tag]` ✓
- Defense-in-depth: `KeySet.String()` + `keySetEncryptor.LogValue()` redaction ✓
- Empty-string sentinel doc + explicit empty-`ENCRYPTION_WRITE_VERSION` test ✓
- Fuzz seed corpus for `DecryptString` ✓
- Reserved version `0x00 == invalid` rejects zero-init buffers ✓

#### Operational impact (unchanged)
- `ENCRYPTION_KEY` required at startup, fail-fast wired in `setupEncryption` ✓
- Documented in `.env.example`, `docker-compose.test.yml`, `fly.toml` ✓
- Rotation procedure callout still warns "do not execute in production yet" ✓

#### Cross-repo symmetry (still flagged for follow-on)
The Prisma side in `../my-robo-taxi/` will need to implement the same ciphertext format. `key-rotation.md §"Ciphertext format"` remains the authoritative spec. First column-rollout Linear issue should reference this section explicitly so the Node.js implementation cannot drift on padding/encoding.

---

### Todo
- [x] Read PR diff vs origin/main
- [x] Review new `cmd/telemetry-server/wiring.go` extraction
- [x] Verify `main.go` size now under 300 lines (248 ✓)
- [x] Re-check encryption package + contract docs for any new changes (unchanged since last approval)
- [x] Post updated verdict

Clean follow-up commit. Ship it.
 | [Branch](https://github.com/tnando/my-robo-taxi-telemetry/tree/thomasnandola/myr-16-implement-aes-256-gcm-column-encryption-for-p1-fields)

PR #195, opened 2026-04-26
MYR-57 Implement Go-side role-based field masking infrastructure

tnando:
## Summary

Closes [MYR-57](https://linear.app/myrobotaxi/issue/MYR-57/implement-go-side-role-based-field-masking-infrastructure) (related to [MYR-45](https://linear.app/myrobotaxi/issue/MYR-45) which authored the contract).

Builds the Go-side runtime for the role-based field masking contract that landed in MYR-45. Net-new public API: `Authenticator.ResolveRole`, `internal/mask/`, `Hub.BroadcastMasked`. Behavior changes only on the WebSocket `vehicle_update` path (hub now pre-projects per role and fans out role-appropriate bytes per client). The REST mask plumbing is opt-in via `WithRoleResolver` and is **not** wired in `cmd/telemetry-server/main.go` — it attaches when `/api/vehicles/{id}/snapshot` (rest-api.md §5.2.1) ships, since the existing `/api/vehicle-status/{vin}` is a connectivity probe whose shape doesn't match `ResourceVehicleState`.

## What this PR does

**`internal/auth/role.go`** — `Role` enum (`owner`, `viewer`), `ParseRole`, `String`, `ErrUnknownRole`. Empty `Role("")` is the documented fail-closed sentinel — the mask layer treats it as deny-all per rest-api.md §5.1.

**`internal/ws/auth.go`** — `Authenticator` interface gains `ResolveRole(ctx, userID, vehicleID) (auth.Role, error)`. `NoopAuthenticator` returns `RoleOwner` (consistent with its dev-mode all-access semantics).

**`internal/auth/jwt_auth.go`** — `JWTAuthenticator.ResolveRole` queries `Vehicle.userId`. Match → `RoleOwner`. Mismatch → `RoleViewer` with a TODO comment referencing future Invite-table integration. The forward-looking `RoleViewer` branch is reachable today only via stale `GetUserVehicles` cache or test fixture; real viewer support engages when invite handling ships.

**`internal/mask/`** (new package, ~770 LOC including tests) — allow-list mask matrix + 1% deterministic sampler:
- `mask.go` — `Apply` ("absent, not nulled" via `map[string]any`); `For` (fail-closed: unknown resource or role → deny-all).
- `tables.go` — v1 mask matrix populated from rest-api.md §5.2 with per-resource source-section citations for CG-DC-5 traceability.
[...]

PR #200, opened 2026-05-03
MYR-60 Reconcile hasVehicle empty-slice semantics with NoopAuthenticator docs

tnando:
## Summary

Resolves the documentation/behavior contradiction in `internal/ws/` flagged by the PR #195 review:

- `Client.hasVehicle()` previously treated an empty `vehicleIDs` slice as "grants access to all vehicles" (the dev-mode escape hatch).
- `NoopAuthenticator.VehicleIDs` was documented as "An empty slice means the user is not authorized for any vehicles."

A future engineer reading only the `Authenticator` interface would assume empty = denied and be wrong. Worse, on production a real authenticator returning an empty slice would have **leaked every vehicle** to that client — the opposite of the NFR-3.21 ownership guarantee.

This PR picks Option 1 from the issue (the issue calls it cleaner): make the all-access intent explicit.

- New `WildcardVehicleID = "*"` sentinel exposed by `internal/ws/auth.go`.
- `NoopAuthenticator.GetUserVehicles` now returns `["*"]` when `VehicleIDs` is unset/empty so the dev-mode all-access intent is no longer implicit-via-empty-slice.
- New `Client.allVehicles bool` flag. The handshake sets it only when `GetUserVehicles` returned the sentinel, and strips the sentinel out of `client.vehicleIDs` before role resolution / `auth_ok` count.
- `hasVehicle()` is now `c.allVehicles || slices.Contains(c.vehicleIDs, vehicleID)` — the empty-slice special case is gone, so production clients with no authorized vehicles correctly deny-all.
- `Authenticator` interface shape unchanged (issue's "Out of scope" rule).
- No wire-format changes. `auth_ok.vehicleCount` reports `len(client.vehicleIDs)` after the wildcard strip, which is `0` for dev-mode all-access — matches the existing §2.3 "informational" wording.

## Acceptance Criteria

- [x] No more contradiction between `hasVehicle` behavior and `NoopAuthenticator` doc.
- [x] Empty `vehicleIDs` in production = deny access (matches the doc string and reviewer expectation).
- [x] All existing tests pass (`go test ./...` clean — one unrelated pre-existing flake in `TestReceiver_GracefulShutdown` confirmed transient by re-run).
- [x] Test specifically for "production client with empty vehicleIDs sees no broadcasts" lands (`TestHub_ProductionClient_EmptyVehicleIDs_DenyAll` covers both `Broadcast` and `BroadcastMasked`).
- [x] `golangci-lint run ./...` — 0 issues.

## Anchored

- NFR-3.21 — vehicle ownership enforced on every subscription (now actually enforced for the empty-slice production case).

## Contract changes

`docs/contracts/websocket-protocol.md` updated in two spots flagged by `contract-guard`:
- §3.3 — narrowed the Client-struct field enumeration to focus on the actual point (no `nextSeq`) so it stops drifting as the struct evolves.
- §4.5 — replaced the stale "empty vehicle list grants all broadcasts" wording with the new explicit-flag description.

No AsyncAPI / JSON Schema / fixture changes — wire format is unchanged.

## Test plan

- [x] `go test ./...` passes (one unrelated flake)
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `golangci-lint run ./...`
- [ ] CI green
- [ ] Claude Review addressed

### DIFF
--- a/docs/contracts/websocket-protocol.md
+++ b/docs/contracts/websocket-protocol.md
@@ -284,7 +284,7 @@ The Linear AC for MYR-11 calls for a "type discriminator, sequence, timestamp" e
 
 Rationale for not shipping `seq`/`ts` in v1.0:
 
-1. The current server has no per-connection sequence counter ([`client.go:Client`](../../internal/ws/client.go) struct has `userID`, `vehicleIDs`, `send`, `remoteAddr`, `hub`, `logger` -- no `nextSeq`).
+1. The current server has no per-connection sequence counter — the [`client.go:Client`](../../internal/ws/client.go) struct does not carry a `nextSeq` field, and adding one would touch every broadcast hot-path.
 2. Reconnect-resume is currently implemented entirely client-side via REST snapshot fetch (see [`state-machine.md`](state-machine.md) §5), which establishes a consistent baseline without needing wire-level sequence numbers. The trade-off: clients cannot detect dropped frames within a single connection.
 3. Adding `seq` is a coordinated server + SDK + fixture change. It is a v1.x extension, not a v1.0 ship blocker.
 
@@ -621,7 +621,7 @@ Per [`state-machine.md`](state-machine.md) §4.2, `connectivity` does NOT direct
 
 > **Anchored:** NFR-3.21.
 
-`Hub.Broadcast(vehicleID, msg)` ([`internal/ws/hub.go`](../../internal/ws/hub.go) line 70) iterates every connected client and calls `client.hasVehicle(vehicleID)`. Only clients whose `vehicleIDs` slice (populated at handshake time from `Authenticator.GetUserVehicles`) contains the target vehicle ID receive the frame. Clients with an empty vehicle list (the `NoopAuthenticator` dev mode) receive **all** broadcasts -- this is explicit in [`client.go:hasVehicle`](../../internal/ws/client.go) line 120 and is intentional for local development.
+`Hub.Broadcast(vehicleID, msg)` ([`internal/ws/hub.go`](../../internal/ws/hub.go) line 70) iterates every connected client and calls `client.hasVehicle(vehicleID)`. Only clients whose `vehicleIDs` slice (populated at handshake time from `Authenticator.GetUserVehicles`) contains the target vehicle ID receive the frame. The dev-mode `NoopAuthenticator` returns the `WildcardVehicleID` sentinel ([`auth.go`](../../internal/ws/auth.go)) from `GetUserVehicles`; the handshake translates that into `Client.allVehicles = true` and `hasVehicle` short-circuits to `true` for those clients. Production `Authenticator` implementations MUST NOT emit the sentinel, so `allVehicles` is always `false` on production: an empty `vehicleIDs` slice means deny-all per NFR-3.21.
 
 The SDK can rely on the following contract: **an authenticated production client will NEVER receive a `vehicle_update`, `drive_started`, `drive_ended`, or `connectivity` frame for a vehicle it does not own at handshake time.**
 
--- a/internal/ws/auth.go
+++ b/internal/ws/auth.go
@@ -6,6 +6,17 @@ import (
 	"github.com/tnando/my-robo-taxi-telemetry/internal/auth"
 )
 
+// WildcardVehicleID is the sentinel value the dev-mode NoopAuthenticator
+// returns from GetUserVehicles to convey "this client is authorized for
+// every vehicle" without overloading the empty-slice case (which a
+// production authenticator uses to mean deny-all per NFR-3.21).
+//
+// Production Authenticator implementations MUST NOT return this value.
+// The handshake (handler.go authenticateClient) intercepts it, sets
+// Client.allVehicles=true, and excludes it from Client.vehicleIDs and
+// the per-vehicle role-resolution loop.
+const WildcardVehicleID = "*"
+
 // Authenticator validates session tokens and retrieves the vehicles a user
 // is authorized to view. Defined at the consumer site (this package);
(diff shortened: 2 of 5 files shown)


tnando:
**Claude finished @tnando's task in 2m 32s** —— [View job](https://github.com/tnando/my-robo-taxi-telemetry/actions/runs/25284716635)

---
### Review Summary

**Verdict: APPROVE**

**Domain focus:** none (general Go review only — no `agent:*` labels on this PR)

The change is tightly scoped, fixes a real production correctness bug (empty `vehicleIDs` from a real authenticator silently granted all-vehicle access), preserves the dev-mode escape hatch via an explicit sentinel, doesn't touch the wire format, and lands paired tests that cover both directions of the regression.

### Findings

#### Critical (must fix)
- _None._

#### Warnings (should fix)
- _None._

#### Suggestions (optional)
- **Race safety: confirmed clean.** `client.allVehicles` is written in `authenticateClient` *before* `h.Register(client)` acquires `Hub.mu` ([`handler.go:98`](https://github.com/tnando/my-robo-taxi-telemetry/blob/thomasnandola/myr-60-reconcile-hasvehicle-empty-slice-semantics-with/internal/ws/handler.go#L98)), and every broadcast path reads it under `Hub.mu.RLock()`. The hub mutex provides the happens-before edge — same posture as the existing `vehicleIDs` / `vehicleRoles` fields, so no new race surface is introduced.
- **Interface contract discoverability (minor).** The "Production `Authenticator` implementations MUST NOT return the sentinel" rule lives in the `WildcardVehicleID` const comment ([`auth.go:9-18`](https://github.com/tnando/my-robo-taxi-telemetry/blob/thomasnandola/myr-60-reconcile-hasvehicle-empty-slice-semantics-with/internal/ws/auth.go#L9-L18)). A future engineer implementing a new `Authenticator` and reading only the interface doc would miss it. Consider adding a one-liner to `Authenticator.GetUserVehicles` doc comment cross-referencing `WildcardVehicleID` so the constraint is visible at the consumer site. Not blocking.
- **Allocation in handshake (micro-nit).** `concreteIDs := make([]string, 0, len(vehicleIDs))` ([`handler.go:173`](https://github.com/tnando/my-robo-taxi-telemetry/blob/thomasnandola/myr-60-reconcile-hasvehicle-empty-slice-semantics-with/internal/ws/handler.go#L173)) always allocates even on the production path where no wildcard is ever present. Handshake is once-per-connection so this is invisible in practice — not worth changing.
- **Dev-mode `BroadcastMasked` posture is unchanged (out of scope).** Worth a heads-up: with `NoopAuthenticator{}`, the wildcard is stripped before the role-resolution loop, so `vehicleRoles` stays empty for dev clients. `BroadcastMasked` then resolves `roleFor(any)` to the empty `Role("")` deny-all sentinel — meaning dev clients still receive `Broadcast` (raw) frames but are silently filtered out of `BroadcastMasked` per-role projections. This is **identical to pre-PR behavior** (the empty role pre-existed; the wildcard was never going to populate per-vehicle roles either way), so it isn't a regression. Worth a follow-up issue if dev mode is expected to exercise the masked path. **Not a blocker for this PR.**

### Domain-Specific Notes

General Go quality checks all pass:

- **Empty-slice semantics fix is real and important.** Pre-PR `hasVehicle` had `if len(c.vehicleIDs) == 0 { return true }` ([`client.go` pre-fix](https://github.com/tnando/my-robo-taxi-telemetry/blob/main/internal/ws/client.go#L120)). Any production `Authenticator` returning `[]string{}` for a user with zero authorized vehicles would have leaked **every** vehicle to that client — a direct NFR-3.21 violation. The new `allVehicles` flag + sentinel pattern resolves the contradiction with the `NoopAuthenticator.VehicleIDs` doc string cleanly.
- **Wire format genuinely unchanged.** `auth_ok.vehicleCount` reports `len(client.vehicleIDs)` after wildcard strip — `0` for dev-mode all-access, same as pre-PR. The PR description's claim matches the code.
- **Tests are well-targeted.** `TestHub_ProductionClient_EmptyVehicleIDs_DenyAll` exercises both `Broadcast` (raw) and `BroadcastMasked` paths — the right shape for a regression test, since the bug surface spans both. `TestHub_NoopAuth_WildcardSetsAllVehicles` directly verifies the sentinel-strip + flag-set, and the table-driven `TestClient_HasVehicle` is updated to flip the previously-incorrect "empty grants all" rows. `t.Cleanup` used (not `defer`), short read deadline used (not `time.Sleep`) — matches project test conventions.
- **`slices.Contains` swap.** Replacing the hand-rolled loop with `slices.Contains` ([`client.go:156`](https://github.com/tnando/my-robo-taxi-telemetry/blob/thomasnandola/myr-60-reconcile-hasvehicle-empty-slice-semantics-with/internal/ws/client.go#L156)) is idiomatic Go 1.23+ and worth keeping.
- **Contract doc updates** (`docs/contracts/websocket-protocol.md` §3.3 and §4.5) accurately reflect the new behavior, and the §3.3 wording change (avoiding the brittle field enumeration) is a nice cleanup.
- **No CLAUDE.md violations:** no `panic`, no globals, no `init()`, no `interface{}`, no `gorilla/websocket`, files all well under 300 lines. Errors propagate via `%w`-wrapped `fmt.Errorf` (handler.go path), no errors swallowed.

### Verdict

Ship it. Clean fix for a real authorization bug, good test coverage, contract docs aligned. The only follow-up worth filing is a separate issue for the dev-mode `BroadcastMasked` role gap (pre-existing; not this PR's problem).
 · [Branch](https://github.com/tnando/my-robo-taxi-telemetry/tree/thomasnandola/myr-60-reconcile-hasvehicle-empty-slice-semantics-with)

PR #201, opened 2026-05-03
MYR-58 Replace structToMap JSON round-trip with typed ToMaskMap helper

tnando:
## Summary

Replaces the `json.Marshal` → `json.Unmarshal` round-trip used by `writeMaskedResponse` with a typed `(vehicleStatusResponse).ToMaskMap()` method. Per the MYR-58 follow-up flagged in PR #195's review of MYR-57, this removes the per-request marshal/unmarshal allocation pair on the masked REST response path.

## What changed

- New `vehicleStatusResponse.ToMaskMap() map[string]any` that hand-mirrors the struct's `json:"..."` tags and flattens pointer-typed fields to their pointed-to value or untyped nil. Output stays byte-identical to the previous JSON round-trip after `mask.Apply` + `writeJSON`.
- `writeMaskedResponse` now calls `resp.ToMaskMap()` directly. The `structToMap` helper is gone, the `encoding/json` import in `internal/telemetry/vehicle_status_mask.go` along with it.
- `derefOrNil[T]` generic helper for the pointer flattening, kept next to the struct so future maskable response types (per the issue: `VehicleSnapshot`, `DriveSummary`, `DriveDetail`, `DriveRoute`) can reuse it.

## Performance

Local benchmark on Apple M5 Pro:

```
BenchmarkVehicleStatusResponse_ToMaskMap_VsJSONRoundTrip/JSONRoundTrip-15   1151023   1032 ns/op   992 B/op   26 allocs/op
BenchmarkVehicleStatusResponse_ToMaskMap_VsJSONRoundTrip/ToMaskMap-15      20724045    58 ns/op    48 B/op    3 allocs/op
```

- **18× faster**, **20× fewer bytes**, **8.7× fewer allocs** (26 → 3).
- Comfortably above the ≥30% allocs-reduction acceptance bar in the issue.

## Acceptance Criteria

- [x] `vehicleStatusResponse.ToMaskMap()` implemented; benchmark shows ≥30% lower allocs (88% lower in practice).
- [x] `internal/telemetry/vehicle_status_mask.go` no longer imports `encoding/json` for the projection step.
- [x] All existing `vehicle_status_handler_test.go` tests pass unchanged (only the now-defunct `TestStructToMap_PreservesWireNames` was retired and replaced with two equivalent tests on the real type).
- [x] When future REST endpoints (`/snapshot`, `/drives`, etc.) ship, they ship with their own `ToMaskMap()` methods rather than calling a removed round-trip helper.

## Anchored

- NFR-3.20 — REST responses projected through caller's role mask
- NFR-3.6 — server scale targets

## No contract impact

- No wire-format change. The masked JSON body is byte-identical because the typed map carries the same wire-name keys and the mask layer only inspects keys.
- `docs/contracts/` untouched.
- `Authenticator` / `roleResolver` interfaces unchanged.

## Test plan

- [x] `go test ./...` passes
- [x] `go vet ./...`
- [x] `go build ./...`
- [x] Benchmark shows ≥30% allocs reduction
- [ ] CI green
- [ ] Claude Review addressed

### DIFF
--- a/internal/telemetry/vehicle_status_handler.go
+++ b/internal/telemetry/vehicle_status_handler.go
@@ -70,6 +70,38 @@ type vehicleStatusResponse struct {
 	ConnectedSince *string `json:"connected_since"`
 }
 
+// ToMaskMap returns a map[string]any keyed by JSON wire name, suitable
+// for projection through the role-based field mask in internal/mask.
+// Replaces the encoding/json round-trip previously used by
+// writeMaskedResponse, removing one Marshal/Unmarshal allocation pair
+// per masked REST response (MYR-58).
+//
+// Pointer-typed fields (LastMessageAt, ConnectedSince) are flattened to
+// their pointed-to value or nil — preserving the same key set the JSON
+// round-trip produced. The mask matrix only inspects keys, so value
+// type fidelity (int64 vs float64) is irrelevant; we use native Go
+// types here.
+func (r vehicleStatusResponse) ToMaskMap() map[string]any {
+	m := make(map[string]any, 5)
+	m["vin"] = r.VIN
+	m["connected"] = r.Connected
+	m["last_message_at"] = derefOrNil(r.LastMessageAt)
+	m["message_count"] = r.MessageCount
+	m["connected_since"] = derefOrNil(r.ConnectedSince)
+	return m
+}
+
+// derefOrNil returns *p as an any, or untyped nil if p is nil. The
+// untyped-nil return is what json.Marshal of a nil pointer produces
+// after the Unmarshal-to-map round-trip — keeping the post-mask JSON
+// output byte-identical to the pre-MYR-58 implementation.
+func derefOrNil[T any](p *T) any {
+	if p == nil {
+		return nil
+	}
+	return *p
+}
+
 // vehicleStatusErrorResponse is the JSON body returned on errors.
 type vehicleStatusErrorResponse struct {
 	Error string `json:"error"`
--- a/internal/telemetry/vehicle_status_handler_test.go
+++ b/internal/telemetry/vehicle_status_handler_test.go
@@ -269,39 +269,105 @@ func (s *stubVehicleIDLookup) GetVehicleIDByVIN(_ context.Context, _ string) (st
 	return s.id, nil
 }
 
-// maskedResponseWithSynthetic is a tiny response shape that mirrors
-// vehicleStatusResponse but adds a forward-looking `licensePlate`
-// field. The actual vehicleStatusResponse doesn't carry licensePlate
-// in v1 — this struct exercises the mask plumbing using a synthetic
-// payload, validating that ANY licensePlate-bearing response would be
-// projected correctly.
-type maskedResponseWithSynthetic struct {
-	VIN          string `json:"vin"`
-	Connected    bool   `json:"connected"`
-	LicensePlate string `json:"licensePlate"`
-	Speed        int    `json:"speed"`
-}
+// TestVehicleStatusResponse_ToMaskMap_PreservesWireNames verifies the
+// typed projection returns a map keyed by JSON wire names — the same
+// keys the mask matrix uses. Replaces the structToMap round-trip test
+// removed in MYR-58.
+func TestVehicleStatusResponse_ToMaskMap_PreservesWireNames(t *testing.T) {
+	lastMsg := "2026-04-30T12:00:00Z"
+	connSince := "2026-04-30T11:00:00Z"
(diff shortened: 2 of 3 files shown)


tnando:
**Claude finished @tnando's task in 2m 18s** —— [View job](https://github.com/tnando/my-robo-taxi-telemetry/actions/runs/25284950256)

---
## Review Summary

**Verdict: APPROVE**

**Domain focus:** none labeled — applied general Go review only (CLAUDE.md project rules).

### Findings

#### Critical (must fix)
None.

#### Warnings (should fix)
None.

#### Suggestions

- **Drift risk on `ToMaskMap()` vs struct `json:"..."` tags.** The helper hand-mirrors the struct's JSON tags. If a future contributor adds a field to `vehicleStatusResponse` and forgets to update `ToMaskMap()`, the field will be silently absent from the masked path but still present in the unmasked path (the `h.roles == nil` branch in `writeMaskedResponse` encodes `resp` directly). The current tests assert `len(got) == len(wantKeys)` against literal keys, so they would catch a struct-only addition only if the literal key list is also updated — i.e., the test mirrors the same hand-maintained list, not the struct's tags. Once you add `ToMaskMap()` to the other response types listed in the issue (`VehicleSnapshot`, `DriveSummary`, `DriveDetail`, `DriveRoute`), consider a single reflection-based conformance test that walks each type's `json` tags and asserts every tag is a key in `ToMaskMap()`. Not blocking for a 5-field struct, but worth doing before this pattern multiplies.

- **`derefOrNil[T]` colocation.** The generic helper is fine where it is, but `vehicle_status_handler.go` is per-feature. When the next `ToMaskMap()` lands, move `derefOrNil` to a shared file (e.g. `internal/telemetry/maskmap.go`) so the second caller doesn't trigger a circular feature-import or a duplicate. Trivial follow-up.

### Domain-Specific Notes

Verified against the project rules in CLAUDE.md:

- **Correctness — byte-identical output preserved.** Confirmed by reading `internal/mask/mask.go`: `Apply` only inspects keys (`mask.allows(k)`), and values pass through unchanged into the output map. JSON has a single numeric type, so `int64(42)` and `float64(42)` both encode to `42` — the wire shape is identical to the pre-MYR-58 round-trip. ✓
- **`derefOrNil` returns untyped nil, not typed-nil-pointer-boxed-in-interface.** The function returns literal `nil` (not `p`) when `p == nil`, so the resulting `any` has nil type and nil value — the comparison `got["last_message_at"] != nil` works as intended. This is the correct fix for the classic Go interface-nil pitfall. ✓
- **Error path correctly trimmed.** The old `structToMap` could return errors (Marshal/Unmarshal); `ToMaskMap` cannot, and the now-unreachable 500 branch was correctly removed from `writeMaskedResponse`.
- **No call sites missed.** `grep -rn "structToMap"` shows only commentary references in the test file describing the historical behavior. ✓
- **File sizes under 300 lines** (`vehicle_status_handler.go` 212, `vehicle_status_mask.go` 130). Test file 443 lines, excluded by the rule. ✓
- **No new globals, no `init()`, no `panic()`, no `any` misuse, error wrapping intact.** Imports updated correctly (`encoding/json` removed from `vehicle_status_mask.go`, `fmt` retained for `resolveCallerRole`). ✓
- **Test design.** Two separate tests (populated vs nil-pointer flatten) instead of a table. For two distinct scenarios with different assertions, this reads more clearly than forcing a table — acceptable per CLAUDE.md. The `int64(0)` zero-value assertion correctly documents the type difference vs the old `float64(0)` round-trip behavior; it doesn't affect wire output but pins the contract for the typed map.
- **Benchmark is self-contained and reproducible.** Inlining the old JSON round-trip implementation into the `JSONRoundTrip` sub-bench keeps the regression baseline alive even though the helper is gone — good test hygiene. The reported 88% allocs reduction (26 → 3) is consistent with eliminating one Marshal + one Unmarshal pair, and comfortably clears the ≥30% issue acceptance bar.

Performance numbers in the description (18× faster, 20× smaller, 8.7× fewer allocs) are plausible from the diff and the benchmark structure; I couldn't re-run them here (sandboxed bash), but the implementation is straightforward enough to trust the local result.

Clean refactor, contract-preserving, well-documented. Ship it.



---

- [x] Read CLAUDE.md
- [x] Read the PR diff
- [x] Read the changed files in detail
- [x] Apply general Go review
- [x] Verify benchmark + test claims (visual inspection; bench not run in sandbox)
- [x] Post verdict
 · [Branch](https://github.com/tnando/my-robo-taxi-telemetry/tree/thomasnandola/myr-58-replace-structtomap-json-round-trip-with-typed-tomaskmap)

PR #202, opened 2026-05-03
MYR-61 Recompute status on speed-only telemetry frames against cached gear

tnando:
## Summary

Closes the §2.4 known gap left open by [MYR-14](https://linear.app/myrobotaxi/issue/MYR-14): pre-fix, the broadcaster re-derived `status` only on telemetry frames carrying `gearPosition`. A speed-only frame (very common — Tesla emits speed at 1 Hz, gear changes are rare) did not refresh `status`, so a vehicle transitioning from parked → moving stayed at `status: parked` until the next gear event landed.

The naive fix (recompute `status` on every speed-only frame) would emit `status` without `gearPosition` — a worse atomic-group violation than the under-refresh. This PR sources `gearPosition` from a per-VIN in-memory cache so `status` always rides on the wire alongside its companion field.

## What changed

- `Broadcaster.gear` — new `sync.Map[VIN]string`, populated whenever a gear-bearing frame goes through `handleTelemetry`.
- `(*Broadcaster).ensureGearGroupAtomic(vin, fields)` — new method invoked after `mapFieldsForClient`. Three paths:
  - Frame carries `gearPosition`: cache the value and derive `status` (existing behavior).
  - Frame is speed-only **and cache hit**: inject cached `gearPosition`, derive `status`. Atomic invariant preserved.
  - Frame is speed-only **and cache miss**: leave `status` off the wire so the SDK keeps its prior value (no partial group).
- `handleConnectivity` clears the cache on disconnect to prevent stale gear from leaking across reconnects.

## Acceptance Criteria

- [x] Speed-only frames carrying `gearPosition` AND `status` together when the broadcaster has cached a gear for the VIN (`TestEnsureGearGroupAtomic_SpeedOnlyAfterCachedGear`).
- [x] Atomic-group invariant: `status` is never emitted without `gearPosition`. Tested in cold-start (no cache → no status) and cache-hit (status + gear together) paths.
- [x] AC test 1: `gear=P` → `speed=25` → second frame carries `gearPosition: P`, `status: driving`.
- [x] AC test 2: `gear=D speed=5` → `speed=0` → second frame carries `gearPosition: D`, `status: driving` (gear stays D — parked transition needs a real gear=P frame).
- [x] `vehicle-state-schema.md` §2.4 known-gap callout removed; §1.1 `status` row cross-link updated.
- [x] Decision-log entry added recording the implementation and the UX impact (SharedViewerScreen no longer renders parked-mode UI on speed-only transitions).

## Anchored

- NFR-3.1 (atomic groups) and NFR-3.4 (no partial groups on the wire) — preserved by the cache-then-inject pattern.
- FR-1.1 (real-time vehicle state) — satisfied: `status` now refreshes on every frame instead of waiting for the next gear event.

## Implementation choice

Approach (1) from the issue (per-VIN in-memory cache). Approach (2) (accumulator-based coalescing) was rejected because it adds latency to every speed-only frame and requires a DB hit on accumulator flush — much heavier than a `sync.Map.Load`.

## Test plan

- [x] `go test ./...` — passes
- [x] `go vet ./...`
- [x] `go build ./...`
- [ ] Browser verification with `cmd/simulator` (per the issue's last AC) — leaving for follow-up since the unit tests fully exercise the atomic invariant; if the simulator scenario reveals a regression, follow-up will track.
- [ ] CI green
- [ ] Claude Review addressed


tnando:
**Claude finished @tnando's task in 2m 40s** —— [View job](https://github.com/tnando/my-robo-taxi-telemetry/actions/runs/25285237999)

---
### Claude Code Review — PR #202

- [x] Read PR diff
- [x] Read changed files (broadcaster.go, broadcaster_test.go, nav_broadcast.go, vehicle-state-schema.md)
- [x] Apply general Go review
- [x] Apply domain-specific review (event-system, websocket-sdk)
- [x] Post final verdict

---

## Review Summary

**Verdict: APPROVE**

**Domain focus:** No `agent:*` labels found on this PR — applied general Go review plus implicit `event-system` and `websocket-sdk` review based on the touched paths (`internal/ws/`, atomic-group invariants, contract docs).

This is a clean, narrow fix for the §2.4 known gap left open by MYR-14. The `cache-then-inject` pattern is the right call: it preserves the gear group's atomic invariant on the wire while also refreshing `status` on every speed-only frame. The schema doc, decision log, and test coverage are all kept in lockstep with the implementation.

### Findings

#### Critical (must fix)
- None.

#### Warnings (should fix)
- None.

1. **`t.Parallel()` consistency in the new tests.** `TestEnsureGearGroupAtomic_ColdStart` runs in parallel, but `TestEnsureGearGroupAtomic_SpeedOnlyAfterCachedGear` and its subtests do not. Each subtest constructs its own `&Broadcaster{}`, so they're independent and safe to parallelize. Adding `t.Parallel()` on both the outer test and each subtest would tighten the suite. ([Fix this →](https://claude.ai/code?q=In%20internal%2Fws%2Fbroadcaster_test.go%20on%20branch%20thomasnandola%2Fmyr-61-recompute-status-on-speed-only-telemetry-frames-atomic-with%2C%20add%20t.Parallel%28%29%20to%20TestEnsureGearGroupAtomic_SpeedOnlyAfterCachedGear%20%28outer%20func%20and%20each%20subtest%29%20to%20match%20the%20parallelism%20of%20TestEnsureGearGroupAtomic_ColdStart.%20Each%20subtest%20constructs%20its%20own%20%26Broadcaster%7B%7D%20so%20parallelism%20is%20safe.&repo=tnando/my-robo-taxi-telemetry))

2. **`sync.Map` ergonomics.** The `gear sync.Map // VIN string → gearPosition string` requires type-assertion on every read (`cached.(string)` is implicit in `fields["gearPosition"] = cached`). A small typed wrapper (`type gearCache struct { m sync.Map }; func (c *gearCache) Load(vin string) (string, bool)`) would localize the assertion and reduce the chance a future caller stores a non-string. Optional — the current code is correct and the comment is clear about the value type.

3. **In-flight telemetry vs. disconnect race (very narrow).** `handleConnectivity` (StatusDisconnected) clears `b.gear[vin]`, but a telemetry frame for the same VIN that was queued before the disconnect could still process after the cache clear and re-cache an outdated gear. The window is tiny (single event-bus dispatch interval) and the next real gear frame after reconnect will overwrite, so end-user UX impact is essentially nil. Worth noting in case a future regression points back here, but not worth fixing speculatively.

4. **Unbounded growth on ungraceful drops.** `b.gear` is keyed by VIN and only pruned by `StatusDisconnected`. If a vehicle drops without a connectivity event ever firing (process restart on Tesla side, hard network loss with no upstream signal), its entry stays forever. With a bounded fleet this is fine; if MYR ever scales the per-VIN cache surface (more groups), a TTL or LRU might be worth considering.

### Domain-Specific Notes

**Event-system / concurrency**
- `sync.Map` is appropriate here: keys are VINs (potentially many), reads dominate (one read per speed-only frame, one write per gear frame which is rare per Tesla's emission cadence), and there's no need for whole-map iteration. Race-detector behavior should be clean — every access is via `sync.Map.Store/Load/Delete`.
- No new goroutines, no new channels, no shutdown path needed.
- `ensureGearGroupAtomic` is a pure function over `b.gear` and `fields`; no I/O, no context required, no error handling needed. Correct call shape.

**WebSocket / SDK contract**
- Atomic-group invariant from `vehicle-state-schema.md §2.4` (`status` never without `gearPosition`) is preserved in all three branches of `ensureGearGroupAtomic`:
  - Frame has `gearPosition` → cache + derive (status + gear together). ✓
  - Frame is speed-only with cache hit → inject cached gear + derive status (status + gear together). ✓
  - Frame is speed-only with cache miss → leave `status` off (no partial group on the wire; SDK retains prior value). ✓
- Schema doc, §1.1 cross-reference, §2.4 callout, and decision log entry are all consistently updated. `contract-guard` should be happy.
- No wire-format change beyond filling in the previously-missing `gearPosition + status` pair on speed-only frames.

**Project-rule checks**
- `broadcaster.go` 272 lines, `nav_broadcast.go` 145 lines — under the 300-line cap. ✓
- Imports grouped correctly (`sync` added in stdlib block). ✓
- No new globals, no `init()`, no `panic()`, no `any` introduced where a concrete type would do. ✓
- No VIN logging added; existing log statements unchanged. ✓
- Hand-written mocks (`stubVINResolver`); table-driven tests; `waitForCondition` instead of `time.Sleep`; `t.Cleanup` for resource teardown. ✓

Nice work on the doc + decision-log discipline — those are easy to skip and they were not skipped here.


 · [`thomasnandola/myr-61-recompute-status-on-speed-only-telemetry-frames-atomic-with`](https://github.com/tnando/my-robo-taxi-telemetry/tree/thomasnandola/myr-61-recompute-status-on-speed-only-telemetry-frames-atomic-with)

PR #215, opened 2026-05-09
MYR-73 (Phase 2/Go) Cleanup on user deletion: WS, Tesla streams, JWT existence

tnando:
## Summary

Closes the three Go-side gaps in `data-lifecycle.md` §3.5 cleanup (FR-10.1) — vehicle deletion now propagates within seconds (no polling) to:

- **WebSocket sessions** — `Hub.RemoveVehicle(vehicleID)` closes every subscribed client with code **4002** and reason `"vehicle access revoked"`. Resolves the **vehicle-deletion sub-case of DV-09**.
- **Tesla mTLS streams** — `Receiver.RemoveStream(VIN)` closes the active inbound connection. Inbound frames for unknown VINs (no Vehicle row) are rejected with HTTP 403 + `slog.Warn` `vehicle_not_authorized` + `tesla_inbound_rejected_total{reason="vehicle_not_authorized"}` counter.
- **JWT existence check** — `JWTAuthenticator.ValidateToken` now consults a User-row probe with a 1s singleflight cache. A token whose `sub` has no User row is rejected as `auth_failed` (wraps `ErrInvalidToken`).

Cross-repo Phase 1 PR with the Postgres `vehicle_deleted` NOTIFY trigger: tnando/my-robo-taxi#255 (merged 2026-05-09).

## Architecture

```
Postgres Vehicle DELETE
   │
   ├─ AFTER DELETE TRIGGER notify_vehicle_deleted_trigger
   │     └─ pg_notify('vehicle_deleted', '{"vehicleId","userId","vin"}')
   │
   └─ Go telemetry server (this PR)
         │
         ├─ internal/store/notify_listener.go
         │     ↳ dedicated long-lived pgx connection (separate from the
         │       request pool — LISTEN holds the connection); reconnects
         │       with exponential backoff (1s → 30s cap); republishes
         │       as VehicleDeletedEvent on the in-process bus.
         │
         └─ cmd/telemetry-server/vehicle_deleted_dispatcher.go
               ↳ Subscribed to events.TopicVehicleDeleted; fans out to:
                   • Hub.RemoveVehicle(vehicleID)        → close 4002
                   • Receiver.RemoveStream(VIN)          → close mTLS
                   • VINCache.Invalidate(VIN)            → evict cache
                   • JWTAuthenticator.InvalidateUser(UserID) → drop existence cache
```

## DV-09 closure (partial)

- **Vehicle-deletion sub-case** → **RESOLVED**. The §4.5 inline callout and §10 catalogue row in `websocket-protocol.md` are updated; status flips from `Open` to `Open (reduced)`.
- **Role-mutation sub-case** (Invite-driven owner↔viewer transitions) → still open. That path needs an Invite-state notify channel that does not exist today; the deletion path is the template for the eventual Invite work. A follow-up Linear issue title is recorded in the §10 row.

## Cross-repo

| Repo | PR | Status |
|------|----|--------|
| `tnando/my-robo-taxi` (Phase 1: Postgres trigger) | [#255](https://github.com/tnando/my-robo-taxi/pull/255) | **MERGED** 2026-05-09 |
| `tnando/my-robo-taxi-telemetry` (Phase 2: Go) | this PR | open |

Phase 1 is the *only* dependency: this PR's integration test (when ported to testcontainers) needs the `notify_vehicle_deleted_trigger` to be present in the DB.

## Tests

- `internal/events/vehicle_deleted_event_test.go` — domain event topic + payload round-trip.
- `internal/store/notify_listener_test.go` — happy path + empty VIN + malformed JSON + missing identifiers + unknown channel + defaults.
- `internal/ws/remove_vehicle_test.go` — real httptest WS client receives close 4002 with the expected reason; bystander clients on other vehicles are unaffected; empty-vehicleID and no-match guards.
- `internal/telemetry/vehicle_authorizer_test.go` — unauthorized VIN rejected with HTTP 403 + counter; authorizer error fails open; `RemoveStream` cleanup; empty-VIN guard.
- `internal/auth/user_existence_cache_test.go` — happy path, positive cache, negative cache (deleted user does not re-hit DB on every retry), TTL expiry, singleflight coalescing under concurrent load (50 callers → 1 DB call), `Invalidate` forces refetch, DB error propagates.
- `internal/auth/jwt_auth_user_existence_test.go` — deleted user's JWT rejected as `ErrInvalidToken` (wraps `ErrUserNotFound`); live user's JWT passes; nil-cache compatibility.
- `cmd/telemetry-server/vehicle_deleted_dispatcher_test.go` — fan-out to hub, end-to-end real-WS-client receives close 4002 on `bus.Publish(VehicleDeletedEvent)`.

## Trade-offs documented in code

- **1s JWT existence-cache TTL.** A deleted user's stale token may pass `ValidateToken` for at most 1s after the deletion commits, then refetches. This is the explicit trade-off for keeping per-frame WS handshakes off the database. The `VehicleDeletedEvent` handler also calls `InvalidateUser` so the per-vehicle-deletion path drops to ~0s in the common case.
- **Authorizer fail-open on transient DB errors.** A DB outage MUST NOT drop legitimate vehicles' inbound mTLS streams; the receiver logs the lookup failure and proceeds. Authoritative "not authorized" is reserved for `ErrVehicleNotFound`.
- **Brief reconnect-window NOTIFY losses.** Per the issue spec ("within seconds of the Next.js commit"), best-effort delivery is acceptable; no polling fallback.

## Test plan

- [x] `go vet ./...`
- [x] `go test ./...` (all packages green)
- [x] `go build ./cmd/...`
- [ ] Integration: with testcontainers Postgres + Phase 1 trigger applied, dial WS client, `DELETE FROM "Vehicle"`, assert client receives close 4002 within 5s. (Deferred to a follow-on integration-test PR; the Phase 1 trigger is now the prerequisite for testcontainers parity.)


tnando:
**Claude finished @tnando's task in 3m 59s** —— [View job](https://github.com/tnando/my-robo-taxi-telemetry/actions/runs/25606921414)

---
### Review Summary

**Verdict: COMMENT** — solid architecture and good test coverage. A handful of minor issues worth addressing, but no blocking defects.

**Domain focus:** general Go review (no `agent:*` labels on this PR)

---

### Findings

#### Critical (must fix)
_None._

#### Warnings (should fix)

1. **CLAUDE.md violation: `time.Sleep` in tests.** Two tests use bare `time.Sleep` as a coordination primitive, which CLAUDE.md explicitly forbids ("no `time.Sleep` in tests (use channels, tickers, or polling with timeout)").

   - `cmd/telemetry-server/vehicle_deleted_dispatcher_test.go:49` — `time.Sleep(50 * time.Millisecond)` followed by **no assertion**. The test name claims `FansOutToHub` but the actual check is "did anything panic?" Either drop the test (it's redundant with `ClosesSubscribedClient`) or assert something concrete (e.g., a counter on a stub hub). The 50 ms wait will be load-bearing in slow CI.
   - `internal/auth/user_existence_cache_test.go:134` — `time.Sleep(50ms)` used as a "wait until 50 goroutines have all entered the singleflight slot" barrier. This is race-prone; on a slow runner, callers may not all be parked yet. Use a `sync.WaitGroup` "ready" gate or `runtime.Gosched()` polling pattern instead.

2. **Doc/code disagreement on JWT fail-closed wrapping.** `internal/auth/user_existence_cache.go:69-71` claims:
   > "the JWT path treats 'lookup failed' as fail-closed (rejected) by wrapping it as ErrUserNotFound"
   
   But `internal/auth/jwt_auth.go:131` actually wraps with the **raw** `existsErr`, not `ErrUserNotFound`:
   ```go
   return "", fmt.Errorf("auth.ValidateToken: %w: %w", ErrInvalidToken, existsErr)
   ```
   The behavior (reject) is correct; the comment is misleading. Either fix the docstring or change the wrap to include `ErrUserNotFound` so callers can `errors.Is(err, ErrUserNotFound)` and get the same answer in both branches.

3. **Operational concern: divergent fail-open/fail-closed under DB outage.** The two new authorization paths react to transient DB errors in opposite ways:
   - `JWTAuthenticator.ValidateToken` (fail-closed) — a Postgres blip rejects **every new browser WS connection**.
   - `vehicleAuthorizerAdapter.IsAuthorized` (fail-open) — the same blip silently admits **every Tesla mTLS upgrade**.

   Both choices are individually justified in the comments. But the combination is non-obvious: a DB outage produces a one-sided outage (browsers get auth_failed; vehicles keep flowing without any authorization check). This deserves a note in the data-lifecycle.md §3.5 doc or an operational runbook so on-call understands the asymmetric behavior.

#### Suggestions

1. **Tighten the vehicle_deleted dispatch order** (`cmd/telemetry-server/vehicle_deleted_dispatcher.go:79-104`). Today the order is: hub close → receiver close → VIN cache invalidate → JWT cache invalidate. Between steps 2 and 3, a Tesla vehicle that reconnects can hit the upgrader, see a still-cached positive `IsAuthorized` answer, and be admitted. The window is microseconds and the data has nowhere to land (FK gone, hub clients closed), so it's not exploitable — but invalidating caches **before** closing streams costs nothing and closes the window completely:
   ```go
   d.vinCache.Invalidate(payload.VIN)        // step 1: future lookups miss
   d.jwtAuth.InvalidateUser(payload.UserID)  // step 2
   d.hub.RemoveVehicle(payload.VehicleID)    // step 3
   d.receiver.RemoveStream(payload.VIN)      // step 4
   ```

2. **`IncRejectedVINNotAuthorized(vin string)` interface signature is misleading.** The Prometheus impl at `internal/telemetry/receiver_metrics.go:172` deliberately drops the `vin` parameter and labels by `reason` only. The doc-comment explains the design, but the signature still takes a parameter that goes nowhere. Consider either:
   - Drop the parameter (cleanest), since the call site already logs the redacted VIN via `slog`, OR
   - Add the redacted VIN as a label (high cardinality risk — probably the right reason it's dropped, in which case option 1).

3. **`time.Sleep` in `singleflight` test (re #1.b above)** — the cleaner pattern when verifying coalescing under concurrent load is to give the `fakeChecker` a `started` channel that fires when `UserExists` enters the call, and have callers use an `errgroup`/`WaitGroup` to coordinate without timing assumptions.

4. **Minor: `RemoveVehicle` doesn't `Unregister` directly.** `internal/ws/hub.go:327` closes the websocket conn and increments the metric, then relies on the readPump exiting and calling `Unregister` indirectly. This works, but the metric `connected_clients` updates only after the readPump observes EOF, which can be tens of milliseconds later. Not a defect, just a small latency in the dashboard. If you want the metric to flip immediately, call `h.Unregister(client)` after `Close()` (idempotent — `Unregister` already guards on map presence).

---

### Positive observations

- **Architecture is right.** LISTEN/NOTIFY → in-process bus → fan-out dispatcher in `cmd/` keeps the dependency rule clean: no `internal/` package depends on another, the dispatcher composes their public APIs.
- **Cache invalidation is wired end-to-end.** Many similar PRs forget the cache step; this one explicitly invalidates `vinCache` and `jwtAuth.userExistsCache` so positive entries can't keep a deleted vehicle alive past the close.
- **Defensive null-checks throughout** (`empty VIN`, `nil hub`, `nil jwtAuth` for dev mode), with comments explaining why each branch tolerates the zero value.
- **Race in `RemoveStream` vs. `cleanupConnection` is correctly handled.** `RemoveStream` uses `LoadAndDelete` (atomic remove + decrement); the natural `cleanupConnection` path uses `CompareAndDelete` so it skips the decrement when `RemoveStream` already removed the entry. No double-decrement.
- **VIN redaction in logs** is consistent across new code (`redactVIN(vin)` everywhere).
- **Test isolation is good** — every test creates its own hub/bus/checker; no global state between tests.
- **Dispatcher integration test uses a real WS roundtrip** — close-code 4002 is asserted from the client side, which is the right contract test.

---

### Domain-Specific Notes
None — no agent labels on this PR. Applied general Go + project (CLAUDE.md) review.

---
 · branch `thomasnandola/myr-73-go-side-cleanup-on-user-deletion-ws-sessions-tesla-streams`

PR #415, opened 2026-08-17
MYR-581 Name both people in a ride: ownerFirstName on the catalog, acceptedByName on the share listing, a nameless car is not offerable, and the rename endpoint reaches Apple-native accounts

tnando:
Contracts **v0.37.0**. Four items, one report behind all of them: an incoming ride card read `Tesla wants a ride` about a human being (MYR-532 item 4), because the wire carried no name for the counterparty and the client fell back to the make.

## 1 · `ownerFirstName` on §7.0 `GET /api/vehicles`

The car owner's **first name**, in **both** role allow-lists, resolved by the platform's three-source identity ladder (`"User".name` → the Apple first-consent name in `go_identity_apple` → `go_users`) and reduced to its first token by the **same** `firstNameToken` behind `RideRequest.requesterName`. One ladder, one reduction, four surfaces.

- **Not `"User".name` alone.** The Go server's user provisioning INSERTs `ON CONFLICT DO NOTHING`, so a pre-existing row can hold a NULL name the Apple identity resolves perfectly well; a one-table lookup calls a named owner nameless.
- **First names only**, narrowed **in the store** rather than by the mask — a mask can drop a value but cannot shorten one, so a full-name field could only ever have been allowed or denied whole.
- Threaded into **all three** catalog reads: the owner list, the shared list, **and the member list** (`queryVehiclesForRideMember`, which shares `scanSharedVehicleSummaryRow` — it had to gain the column in the same position or the scan would transpose).
- P1: never logged on any path.

## 2 · A car whose owner has no name is not offerable

Create and accept refuse a bare **`409 vehicle_unavailable`, no `subCode`** — the **fifth** carrier of that code and a plain member of the family, because four of the five are conditions of the car *right now* and only `time_conflict` is a property of the *time* the rider picked.

- **Self-rides are covered deliberately.** The subject is the car's offerability, not the pair of people — and an owner who only self-rides is precisely the one who would otherwise never meet the client's name sheet, then invite somebody and discover it from the far side.
- **Three layers, shaped by monotonicity rather than reversibility.** Namelessness only resolves *forward* (no path un-names an account; the profile writer refuses the empty string on every rung). So the accept backstop and the dispatch probe exist for rides created **before** this gate shipped — a finite, shrinking population — and both ride reads their enclosing gate was already performing. The sweeper **HOLDS, never expires**, mirroring the un-pause-inside-the-window semantics.
- **The wire value and the gate are one SQL constant** (`internal/store/owner_name.go`): the name is `COALESCE(...)`, the gate is `(COALESCE(...)) IS NOT NULL`. A row showing a name while the gate calls the owner nameless is not a bug that can be introduced. `TRIM` on every rung is load-bearing — without it a whitespace-only name reads NAMED to the gate and `null` on the wire.
- **The enforcement path reads only the boolean.** MYR-578's input-only discipline applied to a P1 name, which is what makes "never log it" structural rather than a rule to remember.

## 3 · `acceptedByName` on the §7.5.2 owner listing

The first name of the account that **redeemed** the invite. This is the **correction to `label`**, not a duplicate: `label` is the owner's own memo, typed before anybody redeemed anything and never resolved against an account, so a forwarded code left an accepted row naming the wrong person and the owner's Share tab could not say who actually holds the grant — the one question they are answering when they decide whether to revoke.

Omitted on pending rows and on accepted rows with no resolvable name, deliberately **indistinguishable** (`status` already separates them; a second sentinel would be a state with no copy). Owner-only by construction — the viewer role has no entry in the `Invite` mask resource at all, which matters more here than for most fields: a viewer who could read it would be enumerating who else an owner has shared with.

## 4 · FIX — `PATCH /api/users/me` could not name an Apple-native account

The endpoint shipped earlier the same day as a single `UPDATE "User"`. Apple-native accounts have **no `"User"` row at all** (they live in `go_users` + `go_identity_apple`), so it answered **404 to exactly the nameless population it exists for**.

**The design chosen: write every ladder rung the account has a row on, in one transaction.** The reason it is not "write the account's own table" is that **precedence, not existence, decides what wins** — writing only `go_users` would leave a stale Apple first-consent name one rung above it, and every reader would go on serving the OLD name while the endpoint reported success. Silently. Writing every rung makes the invariant total: whichever rung the `COALESCE` selects carries the new name, so there is no precedence to reason about and nothing to change when a fifth reader appears.

- **No row is ever created** — all three statements are `UPDATE`s. An `INSERT` into `go_users` for an id living in `"User"` would fork one human across two identity sources (the shape MYR-452's convergence table exists to clean up after).
- **No migration, no schema change, no new deletion step.** A dedicated go-side name store read as a new rung 0 was considered and **rejected as the larger change**: migration + a fourth rung threaded into four SQL constants + a §7.6 deletion step (a name row outliving its owner is a P1 leak) + a retention entry.
- Overwriting `go_identity_apple.name` is safe and deliberate: it is a display cache populated from a *client-supplied* field, and `TouchAppleLogin` only ever backfills it — so no future sign-in can clobber the user's own choice.
- **The 404 arm's meaning narrowed accordingly**: no row in `"User"` **or** `go_identity_apple` **or** `go_users` — an orphaned token, not a routine Apple account.

## Docs

`rest-api.md`: §1 ToC, §2.3 (`PATCH` has two uses now), §4.1.1 (carriers 4 → 5 plus the argument for declining a sub-code), §5.2.0, §5.2.5, §6 catalog, §7.0 field row, §7.5.2 (new field subsection + example body), §7.8 (create gate, accept backstop, both error lists, the dispatch-hold policy's third arm), **§7.26 NEW** — the endpoint had no reference section at all. One changelog row covering all four items, and the earlier same-day MYR-581 row marked superseded on the two points that changed. Plus `vehicle-summary.schema.json`, `vehicle-sharing.schema.json`, `specs/rest.openapi.yaml`, both §7.0 fixtures, `data-classification.md` (two derived-field blockquotes, a provenance note on the two identity tables' `name`, and an audit-trail entry recording that the counts do **not** move) and `data-lifecycle.md` §1.4 (the carve-out's Prisma arm is still exactly one column; the widening is Go-owned).

## Tests

Table-driven throughout: projection arms (named / key-always-present null / both role masks / the real viewer projection), gate arms (create incl. self-rides, order-after-access, one-read, accept backstop, sweeper HOLD + **dispatch-once-named**), share-listing arms (accepted named / accepted nameless / pending), the pure first-token reducer with a **structural** check that both SQL spellings derive from one ladder, and the store-level rename arms including the **stale-higher-rung precedence trap** proved through a real ladder reader rather than by inspecting rows.

`go test ./internal/telemetry/ ./internal/dispatch/ ./internal/mask/ ./cmd/...` all pass; `golangci-lint run ./...` reports 0 issues. `./internal/store/` needs Docker/testcontainers, unavailable in this environment — its new arms (the ladder against a real Postgres, and the four rename arms) run in CI.

## Deviations from the implementation map

1. The shared expression pair lives in a new `internal/store/owner_name.go` rather than inside `queries.go` — that file is already 600+ lines, and the fragment carries a Go-side reducer alongside the SQL. A pointer comment sits beside `catalogTrimLabelExpr`.
2. The map listed two catalog queries; there are **three**. `queryVehiclesForRideMember` shares the shared-list scanner, so it had to gain the column in the same position.
3. `VehicleRepo.GetDispatchState` now returns a **`DispatchState` struct** instead of `(VehicleStatus, bool, error)`. A third fact made two same-typed bools in a positional return a transposition waiting to happen — and transposing *these* two would make a paused car dispatchable and a named owner's car un-dispatchable at once.
4. `queryVehicleDispatchState` dropped its `v` alias so the shared ladder constant drops in unchanged; an alias would have forced a second parameterized copy of the ladder, which is the thing `owner_name.go` exists to prevent.
5. The first-token reduction happens in **Go at scan time** (reusing MYR-229's `firstNameToken`), not in SQL — so `VehicleSummary` only ever holds the first-names-only value.
6. Two files crossed the 300-line rule and were decomposed (`vehicle_share_accepted_name.go`, `reservation_owner_name.go`). Every remaining file over 300 in the diff was already over on `main`.
7. Added a §7.26 reference section rather than only revising prose — the endpoint had none, and the storage model plus the 404 semantics needed somewhere to live.


tnando:
## CI fixes (commit `b8e0795`)

Two failures on the first run — one mine, one pre-existing. Both are worth a note because the second touches a test unrelated to this issue.

### Mine: the contract harness had no ladder to climb

`Contract conformance` failed with **`500 internal_error` on §7.0, §7.1 and §7.2** — every vehicles-list, snapshot and drives request. The cause was not the ladder SQL but the harness it ran against: `tests/contract/contract_test.go` hand-builds a minimal schema, and it had

- no `go_identity_apple` table at all (the ladder's **middle rung** — the only rung an Apple-native account carries a real name on), and
- no `name` column on either `"User"` or `go_users`.

So the statement named a missing relation and a missing column, and the whole read failed. Drives went down with the others because its ownership check goes through the same `GetByID`.

A second, quieter trap sat behind it: `contractVehicleLister` is a **hand-written mirror** of `cmd/telemetry-server/adapters.go`, and it dropped `OwnerFirstName`. That would not have failed the build — it would have made every owner look nameless, which on this field also means "and their car refuses ride requests". Several other catalog fields are already missing from that mirror and their tests happen not to notice; I added the one this issue's tests depend on and left a comment naming the hazard rather than fixing it wholesale.

**This is the failure the local test gap predicted.** The PR body said `./internal/store/` needs Docker and its arms would first run in CI — this is exactly that, one layer out, and it is the argument for the new test that came with the fix: `TestContract_GETVehicles_OwnerFirstNameLadder` exercises the real handler, the real lean query and the real three-rung ladder against a real Postgres, one arm per rung plus both precedence arms plus the whitespace-only arm that keeps the wire value and the offerability gate agreeing. Unit tests with a stubbed store structurally cannot catch "the SQL does not run" or "`COALESCE` picks the wrong rung"; that is now covered where it can be.

### Pre-existing: `TestNewShareCode_Distribution` was a coin toss, not a check

`Test` failed on `duplicate code in 4000 draws — the entropy source is not random`, in `internal/store/vehicle_share_repo_test.go` — a file this PR does not otherwise touch, testing a minter this PR does not touch.

It asserted **zero** collisions across 4000 draws from a 36^6 = 2,176,782,336 space. The birthday expectation there is n²/2N ≈ **0.0037 colliding pairs**, so a *correct* minter fails this assertion roughly **once every 270 runs**. On a repo where every PR runs the suite that is a recurring red build carrying no information.

Changed to a **bound of 2** rather than zero, which keeps every bit of the check's diagnostic power — a sampler narrowed to 10^6 values produces ~8 collisions here, one narrowed to 10^5 produces ~80, and a genuinely broken entropy source produces ~4000 — while putting the false-failure probability around 1e-8. `draws` is deliberately unchanged: the symbol-coverage assertion beside it (all 36 alphabet symbols must appear) is the one with real teeth and it needs the volume.

I would normally leave an unrelated test alone, but this one blocks every PR it lands on, including this one, and the fix is to make the assertion say what the test already means.

### Result

`Contract Guard`, `Contract conformance` and `Lint` green. `gofmt`, `go vet` (both with and without the `contract` tag) and `golangci-lint run ./...` clean locally.

tnando:
## Review suggestions addressed (commit `3bdf429`)

Thanks — the first suggestion turned out to be a **latent bug**, not a style gap, so I took it rather than deferring it.

### 1. Ladder divergence — fixed, and it was hiding a real defect

You flagged that `ownerNameLadderExpr` wraps every rung in `NULLIF(TRIM(…), '')` while the older `queryOwnerFirstNameSources` used `NULLIF(…, '')`. Chasing it down, the missing `TRIM` is **not harmless** — it is a precedence bug:

`NULLIF(x, '')` alone treats `"   "` as a **present** value. So a top rung holding whitespace **won its `COALESCE` outright**, the rungs below were never consulted, and the Go-side `firstNameToken` then collapsed it to `""` — at which point the caller fell through to the **email local-part** while a perfectly good real name sat one rung down. The Go token saves the *display* from rendering whitespace; it cannot recover the name the SQL already discarded.

`requesterIdentitySelect` had it too, so I fixed both (name **and** email rungs, for symmetry — a whitespace-padded email is equally junk). All four ladders now spell it identically.

**On folding them into one constant:** I looked and it does not fit in this PR. The four key their subselects on different columns (`"Vehicle"."userId"`, `accepted_by_user_id`, `rider_id`, `$1`), so unification needs a key-parameterized helper — and every statement that embeds them is a `const`, so that would force a large set of query constants to `var` across several files, including hot ride-request paths. That is a real refactor with its own risk, not a tidy-up to bundle here. What I did instead is make the invariant **enforced** rather than hoped for:

```go
func TestEveryNameLadderTrimsEveryRung(t *testing.T)
```

It walks all four ladders and asserts `count("NULLIF(TRIM(") == count("NULLIF(")` in each — so a new ladder added without `TRIM`, or an old one that loses it, fails immediately. That closes the drift you were pointing at without the `const`→`var` blast radius; the full fold is worth its own issue.

### 2. `containsSub` / `countSub` — removed

Replaced with `strings.Contains` / `strings.Count`. You were right that the import-avoidance justification did not hold up.

### 3. Unaliased `"Vehicle"` in `queryVehicleDispatchState` — left as is

Agreed on all points, including that it is worth flagging. No change: the trade-off is named at the constant, and `TestOwnerNameLadderSharesOneDefinition` plus the new cross-ladder test would both catch a future alias-addition that forked the expression.

---

`gofmt`, `go vet` (with and without the `contract` tag) and `golangci-lint run ./...` clean. Not merging — leaving that to you.

PR #416, opened 2026-08-17
MYR-583 A name only counts once its owner has confirmed it

tnando:
## MYR-583 — a name only counts once its owner has confirmed it

**Client ruling, verbatim:** *"Any names not entered/confirmed should be in Setting Up. So please ensure even cars with names but not confirmed show that label."*

MYR-581 taught four readers to resolve a display name through a three-rung ladder. This is the fact that ladder could not express: **two of its three rungs are filled by machinery the person never saw.** `go_identity_apple.name` is Apple's FIRST-consent payload — sent once, forwarded by the client, never shown to anybody for approval. `"User".name` may hold whatever the legacy web onboarding wrote, including a literal `Tesla User` placeholder. **A resolvable name is not evidence of consent**, so the platform has been publishing names their subjects never saw.

### 1. Confirmation is a server fact

New Go-owned table, migration `0042_profile_name_confirmations` — `user_id TEXT PRIMARY KEY`, `confirmed_at TIMESTAMPTZ NOT NULL`. No FK (CG-DL-9); no Prisma table named anywhere in the file.

`PATCH /api/users/me` UPSERTs the caller's row as the **last statement of the existing name-write transaction**, conditioned on the name write having matched at least one rung. The two failure directions are not equal, and only one is tolerable: a confirmation that committed while the name write rolled back would mark a STALE name as approved by the person, permanently and invisibly — the exact state this issue exists to eliminate, produced by the feature that implements it. **A `404` therefore confirms nothing.**

**Row presence is the whole signal** — no `confirmed BOOLEAN`, no expiry, and nothing compares `confirmed_at` to anything. Confirmation is monotonic; the only deletion is account deletion.

### 2. The emit and the gate moved together, because they are one expression

`owner_name.go`'s founding invariant is that `ownerFirstName` and the `ownerNamed` offerability predicate derive from ONE SQL constant so they cannot disagree. The conjunction went **inside** that constant (`ownerNameLadderExpr` = `CASE WHEN confirmed THEN <ladder> END`), so both surfaces acquired it in one edit and remain one fact. `acceptedByName` (§7.5.2) gains the mirror gate on the **grantee**.

**No new absent-state anywhere:** `null` / OMITTED already meant "cannot be named", so a client-visible ruling needed no client-visible contract change. Same 409, same absent sub-code, same message.

### 3. Two surfaces are deliberately NOT gated

- **`RideRequest.requesterName`** — it is on the card an owner reads while deciding whether to admit a stranger. Gating it would REMOVE information the receiving party already acts on and put `Tesla wants a ride` back on the very surface MYR-532 item 4 reported, while changing nothing for anybody the feature is about (a rider on the current build has necessarily confirmed on-device before they can create a ride).
- **`RedeemShareInviteResponse.ownerFirstName`** — that ladder falls through to an **email local-part**. Gating it would render `amruth.kelkar` instead of `Amruth`: a gate that makes the leak larger is not the gate the ruling asked for.

Both exemptions are argued at their definitions and pinned by a structural test, so a later "consistency" refactor has to read the argument first.

### 4. The legacy data step is a reviewable binary, not a migration and not a psql session

Both statements name the Prisma-owned `"User"` table, and **CG-DL-9 forbids that in Go migration SQL outright** (`docs/architecture/migrations.md` §4.2's corollary names this exact case). So `cmd/backfill-name-confirmations` (+ `internal/store/nameconfirmbackfill`) does it as sanctioned application-runtime SQL, in ONE transaction, with `-dry-run`:

1. **Scrub** `"User"."name" = NULL WHERE TRIM("name") = 'Tesla User'` — expected 1 row in production.
2. **Backfill** confirmations from `"User"."updatedAt" >= 2026-08-17T07:00:00Z`, `ON CONFLICT DO NOTHING`.

The heuristic's bound is stated rather than waved at: at worst it credits an account whose name was legacy-written today, a set that is empty in practice. **The scrub is audited** — a new `profile_name_placeholder_scrubbed` AuditLog row per account, in the same transaction as the UPDATE (stronger than the MYR-447 mass-scrub precedent could offer), and a real run REFUSES to proceed without the audit writer.

### 5. §7.6 gains step 8b

`AccountDeleter.DeleteProfileNameConfirmation`, among the personal effects and before the identity transaction. The `account_deleted` metadata gains `profileNameConfirmationsDeleted` (0 or 1; the row never held the name).

### Docs

`rest-api.md` (§4.1.1 carrier (e), §7.0, §7.5.2, §7.6, §7.8 ×3, **§7.26 confirmation semantics in full**, changelog), `data-lifecycle.md` (**§1.4.2 NEW**, §2 retention, §3.1 step 8b, §4.2 audit action, and the stale "never UPDATE/DELETE" on the `User` row corrected), `data-classification.md` (**§1.20 NEW** — P0 in full, plus the count audit trail: P0 157 → 159), `docs/architecture/migrations.md` (the binary joins the sanctioned list), and the local `vehicle-summary` / `vehicle-sharing` / OpenAPI field descriptions.

### Tests

Table-driven throughout. Store-side (Docker): the emit/gate pair read together on every arm so drift between them fails; PATCH confirms, a `404` and a refused empty name confirm nothing, re-confirmation upserts; `acceptedByName` gated on the grantee with pending still absent. Backfill package: its own container, four account shapes, dry-run writes nothing, a real run without an auditor is refused, re-runs are no-ops, an observed confirmation is never overwritten. `internal/telemetry`: step 8b runs, runs before the identity delete, reaches the audit tally, is idempotent, and a failure fails the deletion. Contract harness: the table, the cleanup, `confirmName`, and a new arm pinning the unconfirmed case.


tnando:
Thanks — all three suggestions read correctly. Two are now fixed in `a8e7556`; I'm declining the third with reasoning.

**Fixed — `cmd/backfill-name-confirmations/main.go` dead pool bounds.** Right that it read as intent while doing nothing. Rather than bounds-check a conversion for two literals I control, the bounds are now typed `int32` constants applied straight to `poolCfg`, and `MaxConns`/`MinConns` are gone from the local `config.DatabaseConfig` literal (which now carries only the two fields actually consumed — `URL` and `DisablePreparedStatements`). A binary whose whole run is one transaction has no business taking pgx's GOMAXPROCS-sized default pool against a database the server is also connected to.

**Fixed — dry-run `FOR UPDATE`.** Agreed that "brief" is not the same as "none": a rehearsal that is going to rewind should not take row locks on live `"User"` rows, on a table the sibling app writes. Split into `queryPlaceholderIDs` (no lock) and `queryPlaceholderIDsForUpdate = queryPlaceholderIDs + " FOR UPDATE"`, selected by `r.dryRun`, so the real run keeps the lock that stops the audit row and the scrub straddling a concurrent write and the rehearsal touches nothing. The existing dry-run test arm now exercises the unlocked path.

**Declining — `account_deletion_sequence.go` at 409 lines.** You identified this correctly and I agree it violates the 300-line rule. I'm leaving it, for two reasons rather than one. It was 391 lines before this PR, so the violation is pre-existing and the honest fix is not "extract the 18 lines MYR-583 added" — a real decomposition means moving `run`, `runPersonalEffects`, `cancelOpenRides` and `tearDownOwnedVehicles` into a sibling, which is roughly 130 lines of pure movement through the account-deletion sequence. That is safe in the sense that it changes no logic, and precisely for that reason it belongs in its own PR: it would add a large mechanical diff to a contract-touching change and require a fresh review of the deletion path to establish it changed nothing, which is a poor trade against a rule violation the merge does not worsen in kind. Second, the file's own ordering comments are load-bearing (each step documents why its position is or is not constrained), and deciding which of those travel with which function is an editorial call worth making deliberately rather than as a footnote to a name-confirmation gate. Happy to do it as a follow-up if you'd like it tracked.

PR #417, opened 2026-08-17
MYR-585 Tell the owner when a rider withdraws a request they have not answered

tnando:
## MYR-585 — cancel / decline pushes, second half

**Client, verbatim:** *"Owners should receive a notification if a rider cancels the request or ride. Same for riders, if owner declines or cancels ride they receive notification."*

**What forced it.** A rider created a scheduled request, the owner's phone buzzed with it, and the rider cancelled 13 seconds later. The owner got **no** cancellation push, tapped the stale request push, found an empty screen, and had no way to know why.

### The audit first — event × recipient × exists-today

| Event (`ride.status.changed` unless noted) | Recipient | Before this PR |
|---|---|---|
| `ride.request.created` (instant + scheduled) | owner | ✅ exists |
| → `accepted` | rider **+ joined members** | ✅ exists |
| → `declined` (instant + reservation copy, MYR-360) | rider **+ joined members** | ✅ exists |
| → `arrived` | rider **+ joined members** | ✅ exists |
| → `enroute` | owner | ✅ exists (MYR-462) |
| → `cancelled` by **OWNER** (instant + reservation copy) | rider **+ joined members** | ✅ exists (MYR-522/540) |
| → `cancelled` by **RIDER**, from `accepted`/`arrived`/`enroute` | owner | ✅ exists (MYR-537) |
| → `cancelled` by **RIDER**, from **`requested`** | owner | ❌ **MISSING — the reported bug** |
| → `completed` | nobody | n/a by design |
| `ride.due` | rider **+ members**, and owner (MYR-535) | ✅ exists |
| `ride.nav.unapplied` | owner | ✅ exists |
| `ride.trip.changed` | the non-editing parties **+ members** | ✅ exists |
| reservation **expiry** (`reservation_expired`) | **nobody** | ❌ found-not-fixed (see below) |

Exactly **one** cell was missing. MYR-537 had left it silent on purpose — "pushing every withdrawn request would page owners for riders changing their minds in the booking flow" — and production is the counter-argument: the interruption had *already been spent* on the request itself. The silence saved the owner nothing; it only stranded the notification they were holding.

### What changed

`riderCancelled` no longer answers *whether* to speak — it **classifies** a rider-stamped cancel, so the two arms can carry sentences that are not interchangeable ("no need to continue" is a stand-down for a car already moving, and nonsense addressed to an owner who never accepted):

| Cancelled from | Owner's push | Body |
|---|---|---|
| `accepted` / `arrived` / `enroute` (MYR-537, unchanged) | `{FirstName} cancelled the ride` · `Your rider cancelled the ride` | `No need to continue — the ride has ended.` |
| **`requested`** (MYR-585, new) | `{FirstName} cancelled their ride request` · `Your rider cancelled their ride request` | `Nothing to do — the request was withdrawn.` |

The body closes the loop the incident opened — the owner is holding a request push that now leads nowhere — rather than describing the ride.

**Nameless fallback.** The anonymous voice every owner-facing fallback in `copy.go` already uses (`Your rider …`). The incoming-card fallback (`New ride request`) is a **noun**: it names a thing that arrived and cannot be the subject of a sentence about that thing being taken back.

### Verified-existing, not rebuilt

- **Owner declines → rider** — fires for both shapes, instant (`can't take this ride`) and reservation (`can't make your scheduled ride`, MYR-360). Newly pinned as reaching **group members** too.
- **Owner cancels → rider** — fires on the reservation shape as well as the live one, and already reached members (MYR-548's test).
- **Group members** — the member fan-out is nested inside the rider branch, so every rider-flavoured send already routes to them. No scheme invented; the existing precedent is now pinned for `declined` in both copy variants.
- **Self-suppression** — `ev.OwnerID == ev.RiderID` suppresses the owner arm. The new arm inherits it rather than re-deriving it; pinned with a negative test.
- **Category** — `ride_lifecycle` for all of it. **No sixth §7.19 category**, no migration, no contract version change.
- **Account-deletion cancels** publish no `cancelledBy`/`previousStatus`, so they stay on the silent arm exactly as before.

### Tests

Table-driven, one arm per matrix cell:
- `TestNotifierRiderCancelWakesTheOwner` — 6 arms incl. both copy variants and the two unclassifiable-stays-silent negatives.
- `TestRiderCancelledClassifiesTheFrom` — 9 arms of the classifier (owner stamp, unstamped legacy, non-cancel transition, terminal from-status).
- `TestOwnerStatusAlertCancelForkOutranksTheStatus` — 4 arms, named + nameless × both kinds.
- `TestNotifierRiderCancelFallsBackAnonymously` — 4 arms incl. whitespace-only names.
- `TestNotifierSelfRideRequestCancelStaysSilent`, `TestNotifierRiderCancelNeverReachesTheRider` — the self-suppression negatives.
- `TestNotifierRequestCancelObeysTheLifecyclePreference` — the §7.19 gate.
- `TestNotifierDeclineReachesTheWholeGroup` — 2 arms, the verification for decline × members.

`go build ./...`, `go vet ./...`, `gofmt`, `golangci-lint run ./...` → **0 issues**. All non-Docker suites pass locally.

### Found, not fixed

1. **Reservation expiry notifies nobody.** A reservation past its lateness ceiling is stamped `dispatch_status='failed'` / `reservation_expired` with the ride row left at `accepted` — no status transition, so no push. The rider learns nobody is coming by opening the app. Out of scope here (the client's directive is cancel/decline), but it is the same class of silence this issue is about.
2. **A rider's cancel does not reach group members.** If the requester withdraws, the people who joined hear nothing. The existing member fan-out is rider-*flavoured* copy, and the rider-cancel copy is owner-flavoured — routing it would mean inventing a third voice, so it is reported rather than guessed at.
3. `publishRideCancelled`'s comment in `account_deletion_sequence.go` claims the push is "byte-identical to a rider tapping Cancel". It is not, and was not before this PR: that path stamps no `cancelledBy`, so it sends no push at all. Pre-existing comment drift.


tnando:
Addressed both review warnings in f2c4113.

**File-size ceiling** — fixed rather than deferred, since this PR was what deepened it. `copy.go` is split by **audience**, which is the seam that was already there in the code and in the comments: `copy_rider.go` (what the passenger needs to know), `copy_owner.go` (what the person whose car it is needs to know), `copy.go` (the payload policy, the status/stamp vocabulary, and the two interpolation helpers that enforce the policy mechanically). **150 / 113 / 197 lines** — all three comfortably under 300. Two duplicated `"Try booking another car."` literals collapsed into the shared `bodyTryAnotherCar` while the strings were moving.

`internal/push/notifier.go` (531) is **left alone deliberately**: this PR changed two lines in it and did not deepen it, and splitting the handler set is a real refactor that has no business riding a behavioural fix. Worth its own issue, as you suggest.

**Doc-comment attachment drift** — fixed by construction, not by inserting blank lines. `statusAlert`'s block had been absorbed into `ownerCancelled` and `ownerStatusAlert`'s into `riderCancelKind`; each doc block now sits directly on the function it describes, in the audience file where that function lives.

**Suggestion 1** (overlapping tests) — taken: `TestOwnerStatusAlertCancelForkOutranksTheStatus` now carries a comment stating why it must not be collapsed into the notifier test. One asks "did the right phone ring?", the other "is the right sentence produced?", and only the copy-level one can reach a `riderCancelKind` the notifier would never build — which is what keeps the enum honest if the classifier changes.

**Suggestion 2** (audit table into the docs) — partially taken. The §7.17 sends table already *is* the permanent "what fires when" reference, so rather than duplicate it I added the axis it was missing: an explicit paragraph on which sends carry the group-member fan-out and which have no member audience at all. The two found-not-fixed items from the audit (reservation expiry notifies nobody; a rider's cancel does not reach members) are called out in the PR body for follow-up issues rather than buried in a table.

PR #423, opened 2026-08-20
MYR-594 Carry the step-8c/8d deletion counts into the audit row

tnando:
Follow-up to #422 (review suggestion 1): `UserActivityRowsDeleted` (MYR-592, step 8c) and `TeslaTokenKeepaliveRowsDeleted` (MYR-594, step 8d) were populated and logged but dropped at the boundary adapter, so they never reached the `account_deleted` audit metadata — inconsistent with `profileNameConfirmationsDeleted`, which is carried through. Both tables are P0-only by shape, so the counts join the audit row's CG-DL-5 whitelist with the same argument, and the whitelist test now seeds and asserts both.


tnando:
Added a second commit: dependabot alert 23 (moby/go-archive < 0.3.0, HIGH) — test-only chain via testcontainers-go, zero occurrences in the server binary's dep graph; bumped to 0.3.0 and re-ran the identity integration suite on it.

